PR #4, opened 2021-08-14
Impl block-run for poem.

BinChengZhao:
Because the thread pool does not implement Sync,
 It cannot be initialized with `lazy_static`.
The Sync is implemented through Mutex wrapping.
To prevent subsequent use of the thread pool from blocking,
A thread pool handle is cloned via a thread-local variable.

`poem` always provides the user-first experience, come and use it ~


BinChengZhao:
The implementation does not fit with the framework design philosophy, and I have turned it off for now.

PR #12, opened 2021-08-21
add shutdown delay

uxgnaw:
waiting on green process to done, avoid client 408 error

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,7 +62,7 @@ tracing = { version = "0.1", optional = true }
 tempfile = { version = "3.2.0", optional = true }
 
 [dev-dependencies]
-tokio = { version = "1.9.0", features = ["rt-multi-thread", "macros"] }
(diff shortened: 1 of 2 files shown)


sunli829:
Thank you for your PR, but I think the `Sever::shutdown_delay` method is not general enough. It seems that it can only use `std::process::exit` to force the program to exit after receiving the interrupt signal.

I think you can refer to `https://docs.rs/warp/0.3.1/warp/struct.Server.html#method.bind_with_graceful_shutdown` to let users customize the shutdown conditions, which is better. 🙂

uxgnaw:
Refactoring it like this:
`#[tokio::main]
async fn main() {
    let app = route().at("/hello/:name", get(hello));
    let server = Server::bind("127.0.0.1:3000").await.unwrap();
    server.
        graceful_shutdown(|| {
            println!("i'm done")
        }).
        run(app).await.unwrap();
}`

PR #13, opened 2021-08-21
a example for json request/response

RustChater:
when json parse failed, return message with json format.


sunli829:
It has been merged manually.

PR #25, opened 2021-09-10
Add example to show how to use cookie & session

lazywalker:
i did it with async_session crate, maybe lack-of-poem but it works :)


sunli829:
Your code is like a poem, thank you very much! 🤣

PR #26, opened 2021-09-11
Html Template impl

lazywalker:
because the poem::web::Template only output plaintext(by the way it use private field which can not be used from outside), if we want to render html from template, we can
```rust
#[handler]
async fn about() -> impl IntoResponse {
    let tpl = AboutTemplate {};
    Html(tpl.render().to_string())
}
```
comparing this one i think a HtmlTemplate impl would be much more user-friendly

```rust
#[handler]
async fn about() -> impl IntoResponse {
    let tpl = AboutTemplate {};
    HtmlTemplate(tpl)
}
```


sunli829:
Thank you so much! 😁

PR #32, opened 2021-09-26
Fix CORS Preflight response

12foo:
Hello! Really liking this library so far, especially the OpenAPI integration. I found a small bug in the CORS middleware:

Allowed origins should be returned in Access-Control-Allow-Origin for preflight responses to work correctly.


sunli829:
Thank you so much!😃

PR #53, opened 2021-10-11
Modify the `Error` type of `FromRequest`

liangyongrui:
Make the response returned by custom error type more flexible.


sunli829:
Thank you very much, these changes are great! 😁

Issue #54, opened 2021-10-11
[poem-openapi] How about put some common attributes on the impl scope?


umaYnit:
## Description of the feature
How about put some common attributes on the impl scope?
such as `prefix_path` and openapi `tag`.

## Code example (if possible)
```rust
#[OpenApi]
#[oai(prefix_path = "/pet", tag = ApiTags::Pet)]
impl Api {
    #[oai(path = "/", method = "post")]
    async fn add_pet(&self, pet: Json<Pet>) -> Result<()> {
        todo!()
    }
// ...
}
```

ollyde:
For anyone else who reads this and can't find docs.

```
#[OpenApi(prefix_path = "/api", tag = "ApiTags::Test")]
impl TestRoutesApi {}
```

PR #56, opened 2021-10-11
Add common attributes for OpenApi. #54

umaYnit:


pymongo:
Looks good to me, I like the common tag attribute

umaYnit:
Wrong operation, will reopen new pr. 😭

PR #81, opened 2021-11-10
Add basic cache for ci

PsiACE:
Setup Swatinem/rust-cache@v1

If the cache is hit, it can reduce the ci time by roughly 8 minutes.

Signed-off-by: Chojan Shang <psiace@outlook.com>


sunli829:
Thanks! 🙂

PR #82, opened 2021-11-11
Fix playground in book

dbanty:
I was just poking around the [book](https://poem.rs/openapi/custom_request.html) and found that pushing the play button results in a compilation error as the default edition is 2015 (which doesn't have `async`). I _think_ this will fix it based on the [mdBook book](https://rust-lang.github.io/mdBook/format/configuration/general.html).

### DIFF
--- a/docs/book.toml
+++ b/docs/book.toml
@@ -4,3 +4,6 @@ description = "Poem Book"
 src = "src"
 language = "en"
 title = "Poem Book"
+
+[rust]
+edition = "2021"


dbanty:
I don't know if running the examples will actually be useful for folks, so maybe disabling the playground altogether is a better move? I was actually trying to see if [this example](https://poem.rs/openapi/custom_request.html) compiles since I think the import is wrong 😁

sunli829:
Thanks! 😁

PR #83, opened 2021-11-11
Ignore redis_session unit test by default.

johnm:
The redis_session unit test presumes that there is a running Redis instance, which is typically not the case. So, marking this as ignored so it doesn't intrude on development flow.


sunli829:
I have fixed it, but thank you for your PR. 😁

PR #88, opened 2021-11-15
Include route to openapi.json in Quickstart.

dbanty:
Personally, the document itself is a very important piece of using OpenAPI. Including it in the Quickstart will hopefully make adoption that tiny bit smoother for new users.

It may be better to put it on its own line like the SwaggerUI with a comment, but I don't know Chinese and figure it's better to keep both up to date 😅

### DIFF
--- a/docs/en/src/openapi/quickstart.md
+++ b/docs/en/src/openapi/quickstart.md
@@ -42,7 +42,12 @@ async fn main() -> Result<(), std::io::Error> {
     // Start the server and specify that the root path of the API is /api, and the path of Swagger UI is /
     poem::Server::new(listener)
         .await?
-        .run(Route::new().nest("/api", api_service).nest("/", ui))
+        .run(
+            Route::new()
+            .nest("/openapi.json", api_service.spec_endpoint())
+            .nest("/api", api_service)
+            .nest("/", ui)
+        )
         .await
 }
 ```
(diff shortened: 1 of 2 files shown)


docs/en/src/openapi/quickstart.md
```diff
@@ -42,7 +42,12 @@ async fn main() -> Result<(), std::io::Error> {
     // Start the server and specify that the root path of the API is /api, and the path of Swagger UI is /
     poem::Server::new(listener)
         .await?
-        .run(Route::new().nest("/api", api_service).nest("/", ui))
+        .run(
+            Route::new()
+            .nest("/openapi.json", api_service.spec_endpoint())
```
sunli829, line 47: You can use `at` here, because `api_service.spec_endpoint` does not contain nested paths. 🙂

```
.at("/openapi.json", api_service.spec_endpoint())
```

sunli829 (commented):
Thank you!

sunli829:
> It may be better to put it on its own line like the SwaggerUI with a comment, but I don't know Chinese and figure it's better to keep both up to date 😅

When you add the English document, I will translate it into Chinese. 🙂

docs/en/src/openapi/quickstart.md
dbanty: Oh interesting, I didn't know about `at`, I'll add that in and it'll serve as a hint to others as well!

dbanty:
Thanks for the quick review @sunli829! Changes have been made with a `TODO:` left in place for you to translate. You should be able to edit this PR to add in the translation directly (if GitHub is to be believed).

sunli829:
Thank you so much! 😁

PR #89, opened 2021-11-15
fix cors error

attliaLin:
fix error: blocked by CORS policy: Reqest header failed content-type is not allowed by Access-Control-Allow-Headers in preflight


sunli829:
Thanks!

PR #93, opened 2021-11-16
Omit empty security schemas from OpenAPI document.

dbanty:
This is mostly because my generator currently interprets the presence of security as a reason to require security (even if it's empty). Really I oughta fix my generator but it looked like security (on the path at least) was intended to be omitted due to [this line](https://github.com/poem-web/poem/blob/04654b8ffd42a7240a08a8059b31c7a3d1d2ef99/poem-openapi/src/registry/mod.rs#L423), so I figured the change would be welcome.

By the way I very rarely touch macros so if there's a more clever way to do this let me know. Thanks!!


sunli829:
Maybe the above changes look clearer, so I still merged them. 🙂

PR #144, opened 2021-12-30
docs: Add some basic documentation to the OpenApi derive macro.

dbanty:
I had a bit of trouble figuring out how to add a prefix to all paths which would reflect properly in the spec/docs UI so I figured I'd add a note to the macro in docs.rs. If there's a better place to stick this info just let me know, happy to move it around.

### DIFF
--- a/poem-openapi-derive/Cargo.toml
+++ b/poem-openapi-derive/Cargo.toml
@@ -29,3 +29,6 @@ indexmap = "1.7.0"
 regex = "1.5.4"
 http = "0.2.5"
 mime = "0.3.16"
+
+[dev-dependencies]
+poem-openapi = { path = "../poem-openapi" }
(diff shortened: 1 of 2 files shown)


dbanty:
Well... just stumbled across [this](https://docs.rs/poem-openapi/1.2.17/poem_openapi/attr.OpenApi.html) which I missed earlier. So clearly this is auto-documented and I just missed it the first time. Maybe instead this should link to that page.

sunli829:
Welcome back! It would be nice if you could move the document here. 😁

dbanty:
> https://github.com/poem-web/poem/blob/master/poem-openapi/src/docs/openapi.md

Yeah it looks like it's already documented there and I just missed it, so I'm gonna close this one.

> Welcome back!

Great to be back! I'm hopefully using this framework for a new API I'm working on, so I'll be around making suggestions and asking questions for the next few weeks most likely 😅

sunli829:
I'm not an OpenAPI expert, so your suggestions are important to me. 😁

PR #148, opened 2021-12-30
docs: Add examples to some OpenApi and Operation attributes.

dbanty:
The specific goal was to add an example of the path param syntax which I had a bit of trouble discovering. ~~I feel like these examples should be tested somehow to stay in sync but not sure the best way to do that.~~

By the way, what happened to the mdbook that used to be around? I found that really helpful 😅 and there's a way in mdbook to pull in examples from docstrings, so the examples are tested, docstrings are present when checking out the docs.rs entry (e.g., where I was originally looking for `#[OpenApi]` docs before my last PR), and everything is in sync.

### DIFF
--- a/poem-openapi/src/docs/openapi.md
+++ b/poem-openapi/src/docs/openapi.md
@@ -2,21 +2,81 @@ Define a OpenAPI.
 
 # Macro parameters
 
-| Attribute   | description                               | Type   | Optional |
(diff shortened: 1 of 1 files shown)


dbanty:
Oh good, `cargo test` is already testing these! Now I just need to make it pass 😬

PR #152, opened 2022-01-05
dep update and do some clean up

attila-lin:


attila-lin:
I will separate this pr

PR #167, opened 2022-01-16
Add HashSet and BTreeSet support to OpenAPI

bbqsrc:
This adds support for HashSet and BTreeSet types to OpenAPI for the array types.

It does not add a custom format or do anything clever.


sunli829:
Thanks! 😁

PR #168, opened 2022-01-16
Add Url support to OpenAPI

bbqsrc:
While `Uri` support exists, the popular crate `Url` is not supported. This implements support for that derived from the `Uri` implementation.


sunli829:
Thanks! 😁

PR #185, opened 2022-01-26
Integrate poem-openapi with `bson::oid::ObjectId`

JoelEllis:
This adds a `Type` impl for `bson`'s `ObjectId` type. 
Some feedback would be greatly appreciated - I'm not sure what `fn raw_element_iter` is supposed to do, or that the name I've given the schema is good.


sunli829:
Thanks! 🙂

PR #187, opened 2022-01-29
Replace `SystemTime` with `Instant`

kitified:
Hey there! Thanks for making such a great library!

I recently got started with Poem and needed to customize the default `Tracing` middleware, so I looked at its implementation and noticed it was using `SystemTime`.

I thought switching to `Instant` would make the code a bit simpler while preserving the intent, but happy to close if you disagree. Thanks!

### DIFF
--- a/poem/src/middleware/tracing_mw.rs
+++ b/poem/src/middleware/tracing_mw.rs
@@ -1,4 +1,4 @@
-use std::time::SystemTime;
+use std::time::Instant;
 
 use tracing::{Instrument, Level};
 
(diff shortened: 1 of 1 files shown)


poem/src/middleware/tracing_mw.rs
```diff
@@ -50,16 +51,12 @@ impl<E: Endpoint> Endpoint for TracingEndpoint<E> {
...
-                (Ok(resp), Err(_)) => {
-                    let resp = resp.into_response();
+                Err(err) => {
                     tracing::info!(
-                        status = %resp.status(),
-                        "response"
+                        error = %err,
+                        duration = ?duration,
+                        "error"
                     );
```
kitified, line 59: This is also a slight change in functionality- I personally thought it'd be useful to see the duration of failed requests too, but totally fine to change this back if you'd prefer.

sunli829:
I agree! Using `SystemTime` will be slightly faster than `Instant`, but keeping the code simple should be more important.

PR #188, opened 2022-01-30
fix: unsupported media-type (415) instead of method not allowed (405)

DDtKey:
Now, if the `ContentType` is incorrect, the server returns a `405` error, but according to the specification, `415` is more appropriate. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415
https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.13

In addition, such a response code is used in most web frameworks 🤔


sunli829:
You are right!👍

PR #191, opened 2022-01-31
[.gitignore] Expand for .vscode and additional sqlite files

kolloch:
Just adding some ignores for files that showed up when testing poem and using Visual Code on it.


sunli829:
Thanks!😁

PR #211, opened 2022-02-22
other small dep update

attila-lin:


poem/src/web/websocket/utils.rs
```diff
@@ -49,6 +49,7 @@ impl From<tokio_tungstenite::tungstenite::Message> for Message {
             Binary(data) => Message::Binary(data),
             Ping(data) => Message::Ping(data),
             Pong(data) => Message::Pong(data),
+            Frame(_) => unimplemented!("frame is not supported"),
```
sunli829, line 52:
```suggestion
            Frame(_) => unreachable!(),
```

attila-lin: Ok

attila-lin:
changed the test because [link](https://github.com/snapview/tungstenite-rs/blob/4e6d339a4e825a737be5fc878c9675b6238a5fe5/src/handshake/client.rs#L113)

sunli829:
Thanks! 😁

PR #221, opened 2022-03-06
Add HTML template for `upload` example

mrxiaozhuox:
Add a HTML page for `upload` example, It makes example use easier!


sunli829:
Thanks!😄

PR #227, opened 2022-03-09
bump regex

attila-lin:
because of [Security advisory for the regex crate (CVE-2022-24713)](https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html)


sunli829:
Thanks! 😁

Issue #214, opened 2022-02-24
Add OpenAPI-support for rust_decimal?


Christoph-AK:
Hey there, 
I would love to directly deserialize incoming API requests into rust_decimal::Decimal without precision loss.

Currently with 

```
use rust_decimal::Decimal;

/// Get recipe resource link
#[derive(Object)]
pub struct GetRecipeResourceLink {
    resource: String,
    relative_amount: Decimal,
}
```

I get `the trait bound 'rust_decimal::Decimal: poem_openapi::types::Type' is not satisfied`.

Can compatibility for this be added? Maybe behind a feature gate, like in [mysql_common](https://docs.rs/mysql_common/latest/mysql_common/#crate-features)?

Christoph-AK:
@sunli829 Thanks so much for the quick implementation!

Two observations:

1. Could you change the serialisation from `to_string` to `[normalize](https://docs.rs/rust_decimal/latest/rust_decimal/struct.Decimal.html#method.normalize)`? This will give cleaner responses. 
2. Less importantly: I wonder if we can somehow make this a json Number instead of a String. I understand that on one hand while initially parsing a json-string into the json value dictionary some loss of precision could already occur, as the json number is most likely represented as float internally, so letting the value stay a string is possibly the only way to keep complete accuracy. On the other hand this way we lose type safety and compatibility with calling js here - annoying. For now with the normalization this should be good enough and keeping the precision is by far most important, but maybe someone will find a better way to represent this. Or maybe make this accept both a string and a number for now, but output as string only?

PR #228, opened 2022-03-09
Openapi: Normalize decimals in responses

Christoph-AK:
Normalize decimals in OpenAPI-responses so for example 0.3 gets returned as "0.3" instead of "0.3000000".


Christoph-AK:
See #214

PR #232, opened 2022-03-11
Implement OpenApi trait for monuples

max-niederman:
Right now, the `OpenApi` trait is implemented on tuples of all lengths from 0 through 16 _except_ 1.

It may seem useless to use a `(T,)` when you could just use `T` directly, but there are some cases where it's useful. For example, I have a macro which generates a function which makes a tuple implementing `OpenApi` from a list of types, but it doesn't work when there's only one type listed.


sunli829:
Any update on this? 🙂

PR #236, opened 2022-03-15
When use Json as an extractor, the `Content-Type` should be `application/[prefix+]json`

zys864:
When use Json as an extractor, the `Content-Type` should be `application/[prefix+]json`, 
if the content-type is not  `application/json`, poem return an error instead of continuing parsing body


sunli829:
LGTM 🙂

PR #242, opened 2022-03-21
Make TokioMetrics endpoint public

Christoph-AK:
I'm using a function to bundle all my routes for consistency in testing etc that depends on the endpoint types being pub.
-> Also consistency with other middlewares.


sunli829:
Sorry, I have changed it. 🙂

Issue #241, opened 2022-03-19
Allow collecting query parameters in a single struct, instead of seperate function parameters in poem-openapi


J00LZ:
This can be done in regular poem using `Request::params<T>()`, but doing so does not generate the required openapi definitions. An extra derive macro that would internally generate the parameters as query parameters could probably work. I'm not sure how much work this would be though...

It seems that the biggest issue here is that a query parameter always binds to a single variable. So even if I make my own `ApiExtractor`, it will only bind to one parameter. 

A different solution to my problem would be allowing you to define (part of) a methods spec yourself, but this also seems only partially doable. Mostly since the way that the `ApiExtractor` currently works allows you to define only a single query parameter.

For the fist one something like this would be nice:
```rust
#[OpenApi]
impl Api {
    #[oai(path = "/foo", method = "get")]
    async fn foo_example(&self, foo: Foo) -> String {
        let a = foo.a;
        foo.b
    }
}

#[derive(FromQuery)]
struct Foo {
    a: i32,
    b: String,
}
```

And for the second one a trait that provides the config maybe?
I'm not exactly sure what's needed, since i'm not that well versed in writing an openapi spec by hand.
```rust
#[OpenApi]
impl Api {
    #[oai(path = "/foo", method = "get")]
    async fn foo_example(&self, foo: Foo) -> PlainText<String> {
        let a = foo.a;
        PlainText(foo.b)
    }
}

#[async_trait::async_trait]
impl<'a> FromRequest<'a> for Foo {
    async fn from_request(req: &'a Request, _body: &mut RequestBody) -> poem::Result<Self> {
        Ok(req.params()?)
    }
}

#[derive(Deserialize)]
struct Foo {
    a: i32,
    b: String,
}

impl CustomData for Foo {
    fn metadata() -> MetaSchema {
        // this should probably be something different...
        todo!()
    }
}
```

If this is more or less impossible to implement then I'll just stick to my method with 9 parameters, which just looks weird but is not a real problem. Thanks in advance!

sunli829:
It's not easy to achieve, but I'll try to do it.

banool:
Just to confirm the request from this PR, for now if I want a struct to be "flattened" such that there is a param for each field, I just have to do that myself and then pack it into a struct?

banool:
@sunli829 Could we re-open this and mark it as "do not delete if stale"? This seems like a great feature idea.

skrcka:
Is this still not possible?

PR #245, opened 2022-03-23
Add the ability for parameters to come from urlencoded forms.

J00LZ:
It only works for a urlencoded body. This works the same as poem's `Form<T>`

This introduces an extra dependency on serde_urlencoded as well, since poem does not reexport it.

I am not 100% sure that this should be in the `payload` module, since you can't really give a form response. 

This is related to #241, even though it does not solve that issue it is enough for me to no longer require that one solved.


sunli829:
This `Form` is very similar to [`Json`](https://github.com/poem-web/poem/blob/afe75efc90257776b7cf1ba5488d8861677cb4f5/poem-openapi/src/payload/json.rs#L16), but encoded with `urlencoded`.

PR #250, opened 2022-04-08
Documentation: Add additional link to FromRequest

Christoph-AK:
I searched the docs for the extractor listings more often then I'm willing to admit, so I'd love to add a link where I would expect it.


sunli829:
Thanks! 🙂

PR #257, opened 2022-04-12
Do not include a body chunk in HTTP 304 responses

### LINKED ISSUE #256: Poem erroneously includes a body chunk for HTTP 304 on HTTP/2
This causes Safari to mark the requests as failed whenever Poem is trying to return 304.

`Static/EmbedFilesEndpoint` basically stops working in Safari over HTTP/2 once the files are cached by the browser.

Eugeny:
Fixes #256

The same bug in Actix: https://github.com/actix/actix-web/pull/2453

### DIFF
--- a/poem/src/error.rs
+++ b/poem/src/error.rs
@@ -35,6 +35,9 @@ pub trait ResponseError {
     where
         Self: StdError + Send + Sync + 'static,
     {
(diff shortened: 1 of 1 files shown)


sunli829:
Thank you so much! 😁

PR #267, opened 2022-04-26
add status log info to error response

xgjhlws:
Error response log shows status code will be more helpful.


xgjhlws:
Is there any progress for this?

sunli829:
Thanks a lot, but this change is not good enough, the `Error::status` method is really useful, so I added it. 🙂

https://github.com/poem-web/poem/commit/3d1d39656178d20bc802e012c0bbf73f6459e391

Issue #260, opened 2022-04-16
poem-openapi doesn't get updated anymore 


glebpom:
After #258 cargo update doesn't longer update `poem-openapi`. The version is always pinned to `1.3.24`

sunli829:
I don't understand, is this caused by some rule of `cargo update`?

glebpom:
It probably happens because other dependencies requier indexmap `1.7`, and choose the latest compatible version of `poem-openapi`

sunli829:
I can't reproduce it, `cargo update` always upgrades `poem` to the latest version.

glebpom:
Create a new project with the following `Cargo.toml`:
```
[package]
name = "test-poem"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
indexmap = "1"
poem-openapi = "1"
```

You will end up with the following `poem` dependencies installed:
```
    Checking poem v1.3.28
    Checking poem-openapi v1.3.24
```

If you remove `indexmap` from `Cargo.toml`, you will see `poem-openapi` gets upgraded:

```
➜  test-poem git:(master) ✗ cargo update
    Updating crates.io index
    Updating hashbrown v0.11.2 -> v0.9.1
    Updating indexmap v1.8.1 -> v1.6.2
    Updating poem-openapi v1.3.24 -> v1.3.28
    Updating poem-openapi-derive v1.3.24 -> v1.3.28
```

Putting it back again and running `cargo update` leads to the following:
```
    Updating crates.io index
    Updating hashbrown v0.9.1 -> v0.11.2
    Updating indexmap v1.6.2 -> v1.8.1
    Updating poem-openapi v1.3.28 -> v1.3.24
    Updating poem-openapi-derive v1.3.28 -> v1.3.24
```

Issue #258, opened 2022-04-13
Consider lowering indexmap version requirement


jamesbirtles:
I've run into https://github.com/tkaitchuck/aHash/issues/95 trying to include poem-openapi into my workspace. The workaround suggested in there is to pin indexmap to 1.6, but poem-openapi-derive requires 1.7. Seeing as 1.7 didn't introduce any new features, i suspect the requirement could be lowered in poem-openapi-derive which would solve this issue for me.

sunli829:
Fixed `v1.3.25` 😁

jamesbirtles:
amazing thanks!

PR #285, opened 2022-06-11
Advance version of indexmap for openapi-derive

banool:
Previously we locked the version of indexmap to `1.6.2` to help a user deal with https://github.com/tkaitchuck/aHash/issues/95 (see https://github.com/poem-web/poem/issues/258). Unfortunately this means that anyone who has up to date dependencies might run into issues preventing them from using this crate, because the deps require `> 1.8`.

For example:
```
$ cargo run
    Blocking waiting for file lock on package cache
    Updating git repository `https://github.com/banool/poem`
    Updating crates.io index
error: failed to select a version for `indexmap`.
    ... required by package `guppy v0.13.0`
    ... which satisfies dependency `guppy = "^0.13.0"` (locked to 0.13.0) of package `determinator v0.8.0`
    ... which satisfies dependency `determinator = "^0.8.0"` (locked to 0.8.0) of package `x v0.1.0 (/Users/dport/a/core/devtools/x)`
versions that meet the requirements `^1.8.0` are: 1.8.2, 1.8.1, 1.8.0

all possible versions conflict with previously selected packages.

  previously selected package `indexmap v1.6.2`
    ... which satisfies dependency `indexmap = "~1.6.2"` of package `poem-openapi-derive v1.3.29`
    ... which satisfies dependency `poem-openapi-derive = "^1.3.29"` of package `poem-openapi v1.3.29`
    ... which satisfies dependency `poem-openapi = "^1.3.29"` of package `aptos-node-checker v0.1.0 (/Users/dport/a/core/ecosystem/node-checker)`

failed to select a version for `indexmap` which could resolve this conflict
```

Pinning the version to `1.6.2` is something the users can do in their own Cargo.toml, the rest of us shouldn't be pinned to an old version by default.

This PR restores the version to the latest, `1.8.2`. Without this PR I can't use poem at all without forking it to make this change.

This will also help address https://github.com/poem-web/poem/issues/260.

### DIFF
--- a/poem-openapi-derive/Cargo.toml
+++ b/poem-openapi-derive/Cargo.toml
@@ -22,7 +22,7 @@ quote = "1.0.9"
 syn = { version = "1.0.77", features = ["full", "visit-mut"] }
 Inflector = "0.11.4"
 thiserror = "1.0.29"
-indexmap = "~1.6.2"                                            # https://github.com/tkaitchuck/aHash/issues/95
+indexmap = "1.8.2"
 regex = "1.5.5"
 http = "0.2.5"
 mime = "0.3.16"


sunli829:
I agree!

banool:
Awesome, what do we need to do to land this?

PR #293, opened 2022-06-15
Add optional support for humantime Duration in poem Object

banool:
I would love to be able to represent durations as part of the OpenAPI spec. As it is now you have to just make a field like "duration_secs" and trust the user will pass the value in as seconds. With this, you can instead get the rich parsing from humantime.

Naturally I made this an optional feature.

Tested with:
```
cargo test --features humantime
```

I see the other external types largely don't have tests so I didn't write specific tests for this either.


sunli829:
Thank you, please use nightly Rust to format the source:

```
rustfmt +nightly --all
```

sunli829:
Thanks 😊

PR #302, opened 2022-07-08
Chose Compress algorithm with priority

### LINKED ISSUE #301: Choose compression algorithms with priority
## Description of the feature
With middleware [Compression](https://docs.rs/poem/latest/poem/middleware/struct.Compression.html), response would be compressed if there is any algorithm specified in `Accept-Encoding` request header.

User agent may specified priority by this header like this: `Accept-Encoding: deflate, gzip;q=1.0, *;q=0.5`. But most browser would not specified priority for now.

For example, Header from Chrome would be like this: `accept-encoding: gzip, deflate, br`. And poem chose first algorithm, which means `gzip`. But usually, `Brotli` would perform better than `Gzip`, actual, all CDN that I know would chose `br` as most as they can.

Maybe we should specified default order for those algorithms, for example, `br gzip deflate`

DCjanus:
close #301

### DIFF
--- a/poem/src/middleware/compression.rs
+++ b/poem/src/middleware/compression.rs
@@ -1,4 +1,4 @@
-use std::{cmp::Reverse, str::FromStr};
+use std::str::FromStr;
 
 use typed_headers::{AcceptEncoding, ContentCoding, HeaderMapExt};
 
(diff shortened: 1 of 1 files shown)


DCjanus:
I can't run test in my computer, hope this would be work

PR #308, opened 2022-07-13
Add optional support for YAML as a general request payload

banool:
This PR adds support for YAML, in all the same ways that there is support for JSON. For the most part I just copied what we do for JSON, but with `serde_yaml`, which is mostly a drop in replacement.

In `poem` itself this is an optional feature, but given we already import `serde_yaml` in `poem-openapi`, I made it just include the YAML payload stuff by default.

Still a work in progress.

### DIFF
--- a/poem-openapi/Cargo.toml
+++ b/poem-openapi/Cargo.toml
@@ -27,6 +27,7 @@ poem = { path = "../poem", version = "1.3.33", features = [
   "tempfile",
   "cookie",
   "sse",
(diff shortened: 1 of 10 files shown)


banool:
I'm going to close this PR, I did some research and it seems like using YAML for payloads is generally frowned upon.

sunli829:
I actually think it's useful, although there aren't many scenarios that use it and it seems to be close to completion?

PR #311, opened 2022-07-13
Add example showing how to add your Content type accept example

banool:


sunli829:
I will fix this example. 🙂

banool:
Awesome thanks a lot, I just couldn't figure out how to accept and return different content types for the same endpoint+status. Looking forward to your updates.

banool:
I'm still iterating on this but I can't get it to work:
```
#[derive(Debug, ApiRequest, Eq, PartialEq)]
pub enum MyRequest<
    J: ToJSON + ParseFromJSON + Send + Sync + Type + ToJSON, B: Type + for<'b> Deserialize<'b>
> {
    #[oai(content_type = "application/json")]
    Json(Json<J>),

    #[oai(content_type = "application/x-bcs")]
    Bcs(Bcs<B>),
}
```

As you can see, I've got different type parameters this time, which as far as I can tell makes this example the same as others, like `poem/poem-openapi/tests/request.rs:17`, so I'm not sure why this still doesn't work, it still only generates the last one in the list:
```
  /:
    get:
      tags:
        - General
      summary: blah
      responses:
        "200":
          description: ""
          content:
            application/x-bcs:
              schema:
                $ref: "#/components/schemas/IndexResponse"
      operationId: blah
```

sunli829:
If you need the same status to return a different content type, you need to use the `ResponseContent` macro.

https://github.com/poem-web/poem/blob/ce664ca8bf4a22dfa3b6a7193a74641f6c3d5ba2/examples/openapi/content-type-accept/src/main.rs#L54

I have fixed the example, can you help me test it? 🙂

banool:
Awesome this works! Thanks so much, it's looking great now.

PR #315, opened 2022-07-14
Demonstrate how to add headers to the response with a custom response type

banool:
Hey, I'm trying to figure out how to add headers to a custom response type. I'm trying to modify an example to demonstrate how to do this. As you can see, I'm trying many different options, but I can't figure out how to do it properly. I see in some other parts of the codebase that you wrap the custom response type with `Response`, but I would like to keep the same return type for `get`, and just inject headers somehow. 

I feel like perhaps if you derive `ApiResponse`, it should expose a function to add headers, just like with the regular `Response` object. Does that make sense? If so, I can take a look at adding support for this.

### DIFF
--- a/examples/openapi/content-type-accept/src/main.rs
+++ b/examples/openapi/content-type-accept/src/main.rs
@@ -59,7 +59,10 @@ enum MyResponseContent<T: ToJSON + Send + Sync + Serialize> {
 #[derive(ApiResponse)]
 enum MyResponse<T: ToJSON + Send + Sync + Serialize> {
     #[oai(status = 200)]
-    Ok(MyResponseContent<T>),
+    Ok(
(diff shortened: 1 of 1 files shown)


sunli829:
You should use `header` attribute to add a header, then it can shown in Swagger.

https://github.com/poem-web/poem/blob/cd27094b82bc07dfaf2cc87690b59a5cb4c7058d/poem-openapi/tests/response.rs#L105

banool:
Oh awesome this makes sense, I'll take it for a spin.

banool:
I updated the example :)

banool:
Hey @sunli829 do you think we could land this? It should be good to go.

sunli829:
Thanks! 🙂

PR #326, opened 2022-07-19
Return 400 when parsing path fails, not 404

banool:
All the other parse errors return 400, but this one returns 404. This PR makes it conform to the other parse errors.


banool:
Hey @sunli829 do you think we could land this? It should be good to go.

sunli829:
Thanks! 🙂

PR #336, opened 2022-07-24
Add support for 128 bit external integers

banool:
Currently it only goes up to 64 bits, but Rust natively supports up to 128 bits, so Poem should probably support it too.


banool:
Disregard, turns out serde doesn't support this.

PR #338, opened 2022-07-24
Add response types related to streaming Bytes

eduardobraun:
I had to implement something similar to this for an internal facade service at work,
and would like to push it upstream.
It creates helper types to use streams of `Bytes` as the response of a handler in `poem` and `poem-openapi`.
Please, feel free to criticize the implementation and naming schemes.

### DIFF
--- a/examples/poem/stream-response/Cargo.toml
+++ b/examples/poem/stream-response/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "example-stream-response"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+futures-util = "0.3.21"
+poem = { path = "../../../poem" }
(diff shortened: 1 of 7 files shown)


sunli829:
I really appreciate your PR, it looks very neat. 👏

However, I think these new codes are unnecessary:

1. `StreamResponse` can be replaced by [Body::from_bytes_stream](https://docs.rs/poem/1.3.35/poem/struct.Body.html#method.from_bytes_stream).

2. `JsonProxyStream` can be replaced by [`actual_type`](https://github.com/poem-web/poem/blob/ec035f0fdce43b80ca9db747d78f81db408b52fb/poem-openapi/tests/api.rs#L675).

eduardobraun:
I didn't know about the `actual_type`, it probably solves my use-case too!
I'm going to close the PR.

PR #340, opened 2022-07-25
Use Parent_Child instead of Parent[Child] for generated intermediate type

banool:
In the OpenAPI specification, it states that these are the only valid characters for a component name:
```
A..Z a..z 0..9 . _ -
```
From https://swagger.io/docs/specification/components/.

As such, Poem doesn't generate conformant component names right now. This PR changes the generation from `Parent[Child]` to `Parent_Child`. I feel this is the best option, `-` and `.` are generally special characters in some languages and that would make the client generation less predictable.

We should bump the minor version for this one since it's a breaking change.


sunli829:
Thanks! 🙂

PR #348, opened 2022-07-27
Use LENGTH_REQUIRED instead of BAD_REQUEST in SizeLimit middleware

banool:
There is a specific error code for this case, let's use it.


sunli829:
You are right!

PR #349, opened 2022-07-27
Add `has_source` function to Error

banool:
This would be helpful to me in the function I'm writing to use with `catch_all_errors`.


sunli829:
Thanks! 🙂

PR #361, opened 2022-08-09
Expose libcookie `iter()`

bvanneerven:
Right now, we cannot iterate over cookies in a cookie jar, because the `iter()` method from the inner `libcookie::CookieJar` is not exposed.
Being able to iterate over the cookie jar is useful whenever we do not know the exact names to `get()` cookies by in advance.
I therefore created this small PR to add the method to the poem API as well.

### DIFF
--- a/poem/src/web/cookie.rs
+++ b/poem/src/web/cookie.rs
@@ -378,6 +378,11 @@ impl CookieJar {
         self.jar.lock().reset_delta();
     }
 
+    /// Returns an iterator over all of the cookies present in this jar.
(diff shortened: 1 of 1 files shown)


bvanneerven:
Since the inner cookie jar is locked behind a mutex, there is not really a way to 'directly' return an iterator; we cannot return a reference to temporary values. 
Although it might not be the best solution (please let me know as I am still learning Rust), the current implementation gets the job done, so I am marking this ready for review :smile:.

sunli829:
Looks great!😃

Issue #365, opened 2022-08-12
`actual_type` is not being inserted on schema registry


eduardobraun:
## Expected Behavior
The `actual_type` attribute should replace the result type on an endpoint.

## Actual Behavior
When using the `actual_type` on a [minimal api](https://github.com/eduardobraun/poem-actual_value-issue-minimal) I get the following runtime error:
```
thread 'main' panicked at 'Schema `MyObj` does not registered', /home/***/.cargo/registry/src/github.com-1ecc6299db9ec823/poem-openapi-2.0.7/src/registry/clean_unused.rs:18:40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

## Steps to Reproduce the Problem

1. Create a new API like [this](https://github.com/eduardobraun/poem-actual_value-issue-minimal)
2. Run `cargo run`

## Specifications

- Version: `poem-openapi = 2.0.7` and `poem = 1.3.37`
- Platform: Linux

## Solution
I've implemented a [fix](https://github.com/eduardobraun/poem/tree/register-actual_type-schema) for this issue, if the issue is confirmed I can make a PR.

sunli829:
Good job, I need this PR. 🙂

PR #366, opened 2022-08-12
Add actual_type to schema registry

eduardobraun:
This PR fixes the issue #365, where the `actual_type` attribute is not being added to the schema registry causing the API server to crash.


sunli829:
Thanks! 🙂

PR #371, opened 2022-08-18
Fix: trace nested route with original uri

brunobell:
As described in title. @sunli829 thanks for you hard work! ❤️


brunobell:
I nested router `/index` on `/api`, now it's tracing as expected as shown below:
```text
2022-08-18T09:44:04.158639Z  INFO request{remote_addr=socket://127.0.0.1:5647 version=HTTP/1.0 method=GET path=/v1/index}: poem::middleware::tracing_mw: response status=200 OK duration=33.8µs
```

sunli829:
Thanks!

PR #384, opened 2022-09-08
Include full error context from anyhow errors in PoemError

banool:
When you just print an anyhow error normally, you only get the top level problem, any context is thrown out. This PR makes it print the full error context: https://docs.rs/anyhow/latest/anyhow/struct.Error.html#display-representations.


sunli829:
Thanks! 🙂

PR #398, opened 2022-09-27
Provide some way to alter only framework errors 

banool:
As it is now, if you downcast an error, the error disappears and the error message becomes "response". This is a start to trying to fix that, though I'm not sure yet if it fixes it completely. The use case is a function like this:
```
// Copyright (c) Aptos
// SPDX-License-Identifier: UNLICENSED

use crate::endpoints::{AptosTapError, AptosTapErrorCode};
use poem::{IntoResponse, Response};

use super::errors::AptosTapErrorResponse;

/// In the OpenAPI spec for this API, we say that every response we return will
/// be a JSON representation of AptosTapError. For our own errors, this is exactly
/// what we do. The problem is the Poem framework does not conform to this
/// format, it can return errors in a different format. The purpose of this
/// function is to catch those errors and convert them to the correct format.
pub async fn convert_error(error: poem::Error) -> impl poem::IntoResponse {
    // In the tap the only error we ever return ourselves is AptosTapError.
    // Therefore, if the error is a different type, it must be an error from
    // the Poem framework.
    match error.downcast::<AptosTapError>() {
        Ok(aptos_tap_error) => {
            println!("aptos_tap_error: {:?}", aptos_tap_error);
            AptosTapErrorResponse::from(aptos_tap_error).into_response()
        }
        Err(poem_error) => {
            let error_string = poem_error.to_string();
            build_error_response(error_string)
        }
    }
}

fn build_error_response(error_string: String) -> Response {
    AptosTapErrorResponse::from(AptosTapError::new_with_error_code(
        error_string,
        AptosTapErrorCode::WebFrameworkError,
    ))
    .into_response()
}
```

### DIFF
--- a/poem/src/error.rs
+++ b/poem/src/error.rs
@@ -394,7 +394,7 @@ impl Error {
             },
             None => Err(Error {
                 as_response,
-                source: None,
+                source: self.source,
                 extensions,
             }),
         }


poem/src/error.rs
```diff
@@ -394,7 +394,7 @@ impl Error {
             },
             None => Err(Error {
                 as_response,
-                source: None,
+                source: self.source,
```
sunli829, line 397: Here `self.source` is definitely `None`.

sunli829:
Does this work for you?

https://github.com/poem-web/poem/blob/d1cc08361144546b8a32982db399abe7ebb54e91/poem-openapi/tests/response.rs#L402

banool:
So I'm doing something like you suggest. First off, I have the middleware I mentioned above configured like this:
```
        let server_future = Server::new(TcpListener::bind((
            self.server_config.listen_address.clone(),
            self.server_config.listen_port,
        )))
        .run(
            Route::new()
                .nest(&self.server_config.api_path_base, api_service)
                .at("/spec.json", spec_json)
                .at("/spec.yaml", spec_yaml)
                .with(cors)
                .catch_all_error(convert_error)
                .around(middleware_log),
        );
```

Then my error code looks like this:
```
/// This is the generic struct we use for all API errors, it contains a string
/// message and a service specific error code.
#[derive(Debug, Clone, Object)]
pub struct AptosTapError {
    /// A message describing the error
    pub message: String,
    /// A code describing the error for programmatic use.
    pub error_code: AptosTapErrorCode,
    /// If we're returning a 403 because we're rejecting the mint request, this
    /// contains additional reasons why.
    pub rejection_reasons: Vec<RejectionReason>,
    /// Submitted transaction hashes, if it got to that point.
    pub txn_hashes: Vec<HashValue>,
    /// How many seconds from now to retry the request, if applicable.
    #[oai(skip)]
    pub retry_after: Option<u64>,
}

impl AptosTapError {
    pub fn new(message: String, error_code: AptosTapErrorCode) -> Self {
        if *USE_HELPFUL_ERRORS.get().unwrap() {
            Self {
                message,
                error_code,
                rejection_reasons: vec![],
                txn_hashes: vec![],
                retry_after: None,
            }
        } else {
            Self {
                message: "hah hah hah".to_string(),
                error_code: AptosTapErrorCode::YeahNahYeahYeahYeahNahYeahNah,
                rejection_reasons: vec![],
                txn_hashes: vec![],
                retry_after: None,
            }
        }
    }

    pub fn new_with_error_code<ErrorType: std::fmt::Display>(
        error: ErrorType,
        error_code: AptosTapErrorCode,
    ) -> AptosTapError {
        Self::new(format!("{:#}", error), error_code)
    }

    pub fn rejection_reasons(mut self, rejection_reasons: Vec<RejectionReason>) -> Self {
        self.rejection_reasons = rejection_reasons;
        self
    }

    pub fn txn_hashes(mut self, txn_hashes: Vec<HashValue>) -> Self {
        self.txn_hashes = txn_hashes;
        self
    }

    pub fn retry_after(mut self, retry_after: u64) -> Self {
        self.retry_after = Some(retry_after);
        self
    }

    // If we're telling people to try again later, it's a 429.
    pub fn status(&self) -> StatusCode {
        if self.retry_after.is_some() {
            return StatusCode::TOO_MANY_REQUESTS;
        }
        self.error_code.status()
    }
}

impl std::fmt::Display for AptosTapError {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        write!(
            f,
            "Error({}): {}: {:?}",
            self.error_code as u32, self.message, self.rejection_reasons
        )
    }
}

impl std::error::Error for AptosTapError {}

// This is only really necessary for the type of the error in the Result
// returned by the endpoint handlers.
#[derive(Debug, ApiResponse)]
pub enum AptosTapErrorResponse {
    Default(
        StatusCode,
        Json<AptosTapError>,
        #[oai(header = "Retry-After")] Option<u64>,
    ),
}

impl From<AptosTapError> for AptosTapErrorResponse {
    fn from(error: AptosTapError) -> Self {
        // We use this opportunity to bump metrics based on the specifics of
        // this response, since this function is only called right when we're
        // about to return this error to the client.
        bump_rejection_reason_counters(&error.rejection_reasons);

        let retry_after = error.retry_after;
        Self::Default(error.status(), Json(error), retry_after)
    }
}
```

The endpoint is then this:
```
    #[oai(path = "/mint", method = "post", operation_id = "mint", hidden)]
    async fn mint(
        &self,
        amount: Query<Option<u64>>,
        auth_key: Query<Option<String>>,
        address: Query<Option<String>>,
        pub_key: Query<Option<String>>,
        // This automagically uses FromRequest to get this data from the request.
        // It takes into things like X-Forwarded-IP and X-Real-IP.
        source_ip: RealIp,
        // Same thing, this uses FromRequest.
        header_map: &HeaderMap,
    ) -> poem::Result<Json<FundResponse>, AptosTapErrorResponse> {
```

I need to do this because the errors returned by the framework are not in the same format as my own errors. 

If I enabled the middleware, the error looks like this:
```
{"error_code":"WebFrameworkError","message":"response","rejection_reasons":[],"txn_hashes":[]}
```

If I disable it, it looks like this:
```
{"error_code":"InvalidRequest","message":"Account address, auth key, or pub key must be provided and valid","rejection_reasons":[],"txn_hashes":[]}
```

The second output is what I want, but it seems like my own errors are not being downcast correctly because the error source is None, instead of my error. I wonder how I make this work?

Ultimately I'm just trying to find a bulletproof way to determine whether an error is from the framework or not, so I can manipulate its format.

banool:
I have a solution that works here (https://github.com/aptos-labs/aptos-core/blob/main/api/src/error_converter.rs) but I think it doesn't work in later versions of Poem (maybe, I haven't confirmed).

banool:
Previously I utilized that my `poem:Error`s didn't have a source but the framework ones do, but this doesn't seem to work anymore. This always felt pretty janky. Could we get either:
1. A new function like `catch_all_error` but instead `catch_framework_errors` that only trigger on errors from Poem.
2. A variable in `poem::Error` like `is_from_framework`.

banool:
I realise this is a big info dump lol but hopefully that last message makes this a bit clearer.

sunli829:
So you actually need to judge which errors are from Poem and which are from Aptos?

sunli829:
Why not do it like below?


```rust

//  ......catch_all_error(convert_error)

impl IntoResponse for AptosError { ... }

async fn convert_error(error: poem::Error) -> impl poem::IntoResponse {
    match error.downcast::<AptosError> {
        Ok(err) => {
            // aptos error
            err.into_response()
        }
        Err(err) => {
            // other errors
            build_error_response(err)
        },
    }
}
```

sunli829:
This is the better solution, I will write an example.

```rust
    #[oai(path = "/mint", method = "post", operation_id = "mint", hidden)]
    async fn mint(
        &self,
        amount: Query<Option<u64>>,
        auth_key: Query<Option<String>>,
        address: Query<Option<String>>,
        pub_key: Query<Option<String>>,
        // This automagically uses FromRequest to get this data from the request.
        // It takes into things like X-Forwarded-IP and X-Real-IP.
        source_ip: RealIp,
        // Same thing, this uses FromRequest.
        header_map: &HeaderMap,
    ) -> poem::Result<Json<FundResponse>, AptosTapErrorResponse> {...}
```

sunli829:
I think I know what's wrong, let me think about it.

sunli829:
I added the `Error::is_from_response` method to determine if the error created from `Error::from_response`. Currently only `openapi::ApiResponse` creates errors with `Error::from_response`, which may help.

banool:
Okay awesome let me try it.

PR #406, opened 2022-10-07
added shuttle as a deployment option to the readme, thanks Sunli <3

ivancernja:
lmk if there are any copy adjustments that need to be made

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,9 @@ The following are cases of community use:
 
 - [Examples](https://github.com/poem-web/poem/tree/master/examples)
 
+### Deployment
+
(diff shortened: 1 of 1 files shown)


Christoph-AK:
To be honest this looks a bit too advertisy for my taste, especially as Shuttle is calling themselves 'not production ready yet'. I would love to see a small tutorial accompanying this, and maybe a change of tone to 'option of deployment' or such. I think it wouldn't be wise to couple this project too tightly to unrelated hosting projects with their own intransparencies.

ivancernja:
@Christoph-AK thanks for the feedback! I've updated the copy a bit to reduce the 'advertisy' appeal and added an example as well. Let me know if this is ok.

PR #434, opened 2022-11-26
fix(poem): not enough randomness in session keys(close #430)

### LINKED ISSUE #430: Not enought randomness in sesssion keys
## Expected Behavior

Generate [32 bytes (256 bits)](https://latacora.singles/2018/04/03/cryptographic-right-answers.html#random-ids) of randomness in `session::server_session::generate_session_id`. This can be achieved by generating 32 pseudo-random bytes and encoding them in base 64 (in url safe mode for this case).

## Actual Behavior

Generate below 24 bytes (192 bits) of randomness (log2(62**32) / 8). 

1. Call `session::server_session::generate_session_id`

## Specifications

- Version: any
- Platform: any
- Subsystem: any

attila-lin:


sunli829:
Thanks!😄

PR #450, opened 2022-12-16
🎨: Support fall back to the index file when serving static files

afitzek:
This adds an option in the StaticFilesEndpoint that can be used to fall back onto the index file, if the static files don't exist. This is useful for hosting SPAs, that are developed with Angular, React, etc.

Change-type: minor
Signed-off-by: Andreas Fitzek <andreas.fitzek@gmail.com>


sunli829:
Thanks! 🙂

Issue #455, opened 2022-12-19
support elliptic curve private keys with rustls


miquels:
## Expected Behavior
Support EC private keys that start with "-----BEGIN EC PRIVATE KEY-----"

## Actual Behavior
ERROR poem::listener::rustls: invalid tls config. error=failed to parse tls private keys

## Steps to Reproduce the Problem

1. Try to load, with rustls, a private key that starts with "-----BEGIN EC PRIVATE KEY-----"
2. Server fails to start

## Specifications

- Version: 1.3.50
- Platform: FreeBSD
- Subsystem: 

I have fixed this bug already, here: https://github.com/miquels/poem/commit/c557044bf5e775a413d5e41f538963c5f386810a

It would be great if you could add this to the next release. Thanks!

FallingSnow:
Can this be closed?

miquels:
fixed by https://github.com/poem-web/poem/pull/460

PR #460, opened 2022-12-20
listener::rustls: add support for elliptic curve private keys

miquels:
See also https://github.com/poem-web/poem/issues/455


sunli829:
Thanks! 🙂

miquels:
Do you want me to fix those rustfmt errors, or is that no problem ?

sunli829:
> Do you want me to fix those rustfmt errors, or is that no problem ?

yes, format the code with `cargo +nightly fmt` 🙂

Issue #461, opened 2022-12-21
Abstract HTTP Path Pattern in Metrics


kolloch:
## Description of the feature

For the request metrics, I see two labels that roughly do what I want: `http.url` and `http.path_pattern`.

![image](https://user-images.githubusercontent.com/339354/208863729-f3e27ee7-8277-4250-906b-2e3e2c11665b.png)

`http.url` is to specific and includes the concrete URL including IDs "resolved" in the pattern. I cannot use it, since it will increase the cardinality of the metric drastically (and therefore increase our bills too much). Basically, a separate time series is stored for all label combinations in most metric systems that I know.

The `http.path_pattern` sounds like what I want but only includes the route prefix at the point where I mounted the middleware, not of the nested route.

The corresponding watched/subscriber routes are mounting like this:

```rust
Route::new().nest(
                    opt.public_base_url.path(), // = /private/nx-release/control/
                    Route::new()
                        .nest("/api", api_service)
                        .at("badges/subscriber/:subscription_id", get(subscriber_badge))
                        .at("badges/watched/:git_ref_id", get(watched_badge))
                        .at("request_dump", post(request_dump))
                        .nest("/", ui)
                        .data(opt.clone())
                        // ... more data(..) ...
                        .data(callback_server.clone())
                        .with(poem::middleware::Tracing)
                        .with(middleware::AddRequestId)
                        .with(OpenTelemetryMetrics::new()) // <- adding the instrumentation to capture all routes
```

I'd love `/private/nx-release/control/badges/subscriber/:subscription_id` as the `http_pattern` (or another label).

## Existing Code

This is were the `http.path_pattern` is added.

https://github.com/poem-web/poem/blob/master/poem/src/middleware/opentelemetry_metrics.rs#L82-L85

```rust
        if let Some(path_pattern) = req.data::<PathPattern>() {
            const HTTP_PATH_PATTERN: Key = Key::from_static_str("http.path_pattern");
            labels.push(HTTP_PATH_PATTERN.string(path_pattern.0.to_string()));
        }
```

`PathPattern` is a simple string container:

```rust
#[derive(Debug)]
#[allow(unreachable_pub)]
pub struct PathPattern(pub Arc<str>);
```

It is set here:

https://github.com/poem-web/poem/blob/4692e5bb6b3422f05cf9df6a7679ea1c1c005ea9/poem/src/route/router.rs#L319-L328

```rust
                match (req.data::<PathPattern>(), req.data::<PathPrefix>()) {
[...]

PR #462, opened 2022-12-21
[PathPattern] Record and use PathPattern in response

kolloch:
...so metrics correctly report the innermost PathPattern.

https://github.com/poem-web/poem/issues/461


poem/src/middleware/tracing_mw.rs
```diff
@@ -47,7 +47,7 @@ impl<E: Endpoint> Endpoint for TracingEndpoint<E> {
         );
 
         if let Some(path_pattern) = req.data::<PathPattern>() {
-            span.record("path_pattern", path_pattern.0.as_ref());
+            span.record("path_pattern", &path_pattern.0.as_ref());
```
kolloch, line 50: @sunli829 This was necessary to make poem compile locally. Not sure why it passed in CI/CD in the old version.

I compiled with rust 1.65

sunli829:
This is a quality PR, thanks a lot! 🙂

Issue #470, opened 2022-12-28
Provide optional URL prefix to generated spec json


VersBinarii:
## Description of the feature
Would you consider adding an option to pass in url prefix into the generated spec?
```
"paths" : {
    "/my_prefix/rest/of/my/path" : {
      "get" : {
        "parameters" : [
          {
            "name" : "X-Customheader",
            "in" : "header",
            "required" : true,
            "schema" : {
              "type" : "string"
             }
           }
         ]
      }
    ... snip ...
 
```

## Code example (if possible)
Maybe something like that:

```
let spec_options = SpecOptions::default().with_url_prefix("/my_prefix");
let my_api_schema = my_api.spec_endpoint_with_options(options);
```

sunli829:
Maybe it can be solved with `prefix_path` attribute.

https://github.com/poem-web/poem/blob/7ada5b69c94c0a98ddff64172703ed61c3386fdf/poem-openapi/tests/api.rs#L88

VersBinarii:
Unfortunately not. This prefix comes from config file and additionally we have abstraction over the handlers so the prefix is dynamic.

sunli829:
Welcome to PR😁

PR #473, opened 2023-01-03
Allow optional prefix in generated spec

VersBinarii:
This PR addresses this [issue](https://github.com/poem-web/poem/issues/470).

@sunli829  please review and let me know if this is ok with you


VersBinarii:
The tests are failing because some rate limits while installing protoc

VersBinarii:
@sunli829 if you're happy with the approach i could add the options to the `swagger_ui` and `rapidoc` functions as well.

sunli829:
Maybe it's better to do this?

```rust
 let api_service = OpenApiService::new(TodosApi, "Todos", "1.0.0")
                              .server("http://localhost:3000")
                              .url_prefix("/foo/bar"); // specified url prefix
```

VersBinarii:
Very well, that seems cleaner. Let me have a look at it then

VersBinarii:
Hello @sunli829 , i've updated it as per your suggestion.

sunli829:
Thanks! 🙂

PR #477, opened 2023-01-10
Setup dependabot to keep dependencies up to date

itsbalamurali:
Set up dependabot to keep dependencies up to date.


sunli829:
thanks

PR #486, opened 2023-01-15
Fix html typo in auth example

denishamann:
The html was probably copy pasted from the CSRF example but the `auth` example has nothing to do with CSRF so I adjusted the html titles accordingly.


sunli829:
Thanks! 😂

PR #494, opened 2023-01-29
Fix #493: Expose ACME structures for user-managed certificate renewal

### LINKED ISSUE #493: Acme cert hook
## Description of the feature
I need to synchronize a certificate across multiple servers.  This means basically
1. Store the acme cert in a database
2. Prevent renewal if renewal is already happening on another server, and probably also store the challenge keys in the database

AFAICT there's no way to do this, the closest I can get is watching the cache directory for changes but this involves extra code and failure cases and could lead to duplicate renewals.

I think concretely, making `issue_cert()` public would be great.

I might have missed something obvious again (maybe this is already possible).

## Code example (if possible)
This is very rough, but something like
```rust
let listener = TcpListener::bind("127.0.0.1:3000").rustls(async_stream::stream! {
    loop {
        let cert = {
            let txn = db.transaction().await?;
            let mut cert = db_lock_cert(&txn).await?;
            if cert.almost_expired() {
                let renewer = auto_cert::new();
                let challenge = renewer.prep_acme().await?;
                db_set_challenge(&db, challenge).await?;
                cert = renewer.finish_acme().await?;
                db_store_cert(&txn, cert);
            }
            cert
        };
        let next_expiry = cert.next_expiry();
        yield cert;
        sleep_until(next_expiry).await;
    }
});
```

andrewbaxter:
Sorry for plowing ahead with this, but I was blocked on it in another project and wanted to make some progress.

Basically this exposes `issue_cert` along with the surrounding necessary structures, methods, etc. I made the following big changes:

* `issue_cert` was made public and its responsibilities were restricted to _only_ completing the ACME procedure.  Using the certificate is left to the caller (no serializing, no updating the resolver, etc).
* A new simplified (no caching, no `spawn`ed TLS update loop) 443 listener was added
* The Http01 port 80 listener was made public

The following shared state structures were also made public
* `AcmeClient` - used by `issue_cert` but may be shared across invocations
* `ResolveServerCert` - used by `issue_cert`, `issue_cert` caller, and 443 listener
* http01 key map - used by `issue_cert`, 80 listener

And some other helper functionality was exposed, methods for avoiding transitive deps, etc.

There's an example showing the usage (just showing the parallel to the fully poem-managed ACME http01 process).  I believe it's minimal: there's not a lot of boilerplate and exposed functions and objects have clear uses, etc.

I'm happy to rename/reorg things, improve comments if things are unclear, or if there are other changes needed.


andrewbaxter:
`key_pair` was only used by `issue_cert` and was already accessible via the `AcmeClient` so I changed `issue_cert` to access it that way.  Otherwise more internal structures would need to be exposed and passed through multiple layers, and I couldn't think of other ways that this would be needed.

sunli829:
I’m so sorry for taking so long to handle this PR because I’ve been busy with other things. 🙂

andrewbaxter:
Oh wow, thanks! No worries, and I'm glad you thought it was reasonable!

PR #498, opened 2023-02-03
fix errors when parse yaml & xml request

liberwang1013:
- add `ParseFromYAML` & `ParseFromXML` implements for  derive macro `Object`
- fix implement errors in `ParsePayload` for and & yaml


attila-lin:
Nice catch!

Would you please add some test case for this?

liberwang1013:
> Nice catch!
> 
> Would you please add some test case for this?

@attila-lin ok, I will summit some test cases in next commit

attila-lin:
Thanks!

PR #502, opened 2023-02-08
fix: make more conflict deps to workspace, pin `quick-xml`

attila-lin:
Related issue: https://github.com/tafia/quick-xml/issues/540


attila-lin:
maybe add to CI latter.

```sh
RUSTFLAGS="$RUSTFLAGS -D unused-crate-dependencies" cargo build
```

attila-lin:
Don't know why.

Dig later.

PR #504, opened 2023-02-08
fix(openapi): increase recursion limit

DDtKey:
`poem-openapi` it started  to fail under `cargo-tarpaulin` with `recursion limit`.

Example of failed CI job could be found in `poem` itself: https://github.com/poem-web/poem/actions/runs/4120444302/jobs/7115175776

### DIFF
--- a/poem-openapi/src/lib.rs
+++ b/poem-openapi/src/lib.rs
@@ -113,6 +113,7 @@
 #![deny(private_in_public, unreachable_pub)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![warn(missing_docs)]
+#![recursion_limit = "512"]
 
 /// Macros to help with building custom payload types.
 #[macro_use]


poem-openapi/src/lib.rs
```diff
@@ -113,6 +113,7 @@
 #![deny(private_in_public, unreachable_pub)]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 #![warn(missing_docs)]
+#![recursion_limit = "256"]
```
DDtKey, line 116: In general, it could be under `cfg_attr`, like: 
`#![cfg_attr(tarpaulin, recursion_limit = "256")]`

However, I see no issues to keep it as general limit, because it doesn't affect result binary, but only compile time.

DDtKey:
It doesn't help, even `1024` limit. 
However it works fine with `2.0.23` version 🤔  

@sunli829 do you have any ideas? I don't think it's directly related to `poem-openapi`, because it happens only under `tarpaulin` (and it has his own instrumentation).

PR #505, opened 2023-02-08
Fix #500: Webpki roots

### LINKED ISSUE #500: Webpki roots for acme client
## Description of the feature
When static linking native TLS roots don't work, or may not be available.  I need to use webpki roots for this. Right now the acme client is hardcoded with
```
                .with_native_roots()
```

Other libraries expose features for each of the 3 roots types (native, manual, and webpki) but alternatively maybe AcmeClient could take an http client as a parameter.

## Code example (if possible)

andrewbaxter:
This works to resolve CA issues in static binaries.

I added two new features:
- `acme-native-roots` which `acme` aliases for backwards compatibility.  I left `acme` in the documentation though, maybe I should replace that with `acme-native-roots`?
- `acme-webpki-roots` which selects webpki roots

I think this is similar to what other libraries do.  I thought this was clear and would provide a path for doing manual roots too in the future.  However, if you don't select either feature, there will be a compile error when building the AcmeClient.

An alternative is to have `acme-webpki-roots` be a toggle, where if it's missing native roots are used and if it's enabled native roots are disabled (and webpki roots are enabled).  This restricts it to exactly two modes however.

### DIFF
--- a/poem/Cargo.toml
+++ b/poem/Cargo.toml
@@ -48,7 +48,10 @@ i18n = [
     "unic-langid",
     "intl-memoizer",
 ]
-acme = [
+acme = ["acme-native-roots"]
+acme-native-roots = ["acme-base", "hyper-rustls/native-tokio"]
+acme-webpki-roots = ["acme-base", "hyper-rustls/webpki-tokio"]
+acme-base = [
     "server",
     "hyper/client",
     "rustls",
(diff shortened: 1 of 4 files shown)


sunli829:
Thanks a lot, I replaced `acme` in the docs with `acme-native-roots`.

andrewbaxter:
Awesome, thank you!

PR #507, opened 2023-02-08
Add From<HashMap> implementation for I18NArgs

Galitan-dev:
I think the tuple way of mapping I18NArgs is very *poor*. For instance, when you don't know the amount of args, you must **handle each amounts**. So I implemented From\<HashMap>.

My personal use case is making a filter `translate` for tera templating:
```jsx
<h1>{{ "hello" | translate(name=name) }}</h1>
```

```rust
fn filter(&self, id: &Value, args: &HashMap<String, Value>) -> Result<Value> {
    self.locale
        .text_with_args(id.as_str().unwrap(), args)
        .map(|str| Value::String(str))
        .map_err(|err| Error::msg(err))
}
```

### DIFF
--- a/examples/poem/i18n/src/main.rs
+++ b/examples/poem/i18n/src/main.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use poem::{
     get, handler,
     i18n::{I18NResources, Locale},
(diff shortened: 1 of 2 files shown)


Galitan-dev:
Let me know, *fellow poem users*, if you want me to implement this i18n tera filter __directly__ into poem.

sunli829:
You are right, I need this.

Issue #512, opened 2023-02-17
Support Tera Templating


Galitan-dev:
# Support Tera Templating

## Description

Adding a Middleware which injects a Tera instance to each request (like Data) to avoid using lazy_static as the current example does and adding a wrapper for tera::Result to avoid the redundant mapping.

## Usage Example

```rust
#[handler]
fn hello(tera: Tera, Path(name): Path<String>) -> tera::Result {
    let mut context = Context::new();
    context.insert("name", &name);
    tera.render("index.html.tera", &context)
}

#[tokio::main]
async fn main() -> Result<(), std::io::Error> {
    let app = Route::new()
        .at("/hello/:name", get(hello))
        .with(poem::TeraTemplating::from_glob("templates/**/*"));
        
    Server::new(TcpListener::bind("127.0.0.1:3000"))
        .run(app)
        .await
}
```

## Future Improvement Ideas 

- implement TeraTemplating::custom(tera: Tera) -> TeraTemplating
- implement a tera i18n filter as mentioned in #507
- add a layer over tera:
```rust
#[handler]
fn hello(Path(name): Path<String>) -> poem::Template {
    poem::Template::render("hello.html.tera")
        .with("name", name)
}
```
- support other template engines like handlebars and liquid

Galitan-dev:
I'm already planning to implement it, but I wanted to let you know so you can share your opinion.
Also, writing an issue helps me make my thoughts clearer in my head 😅.

Galitan-dev:
In case you want to monitor progress:
https://github.com/Galitan-dev/poem/tree/feat/tera

Galitan-dev:
The example is already 13 lines shorter!

```rust
use poem::{
    get, handler,
    listener::TcpListener,
    web::Path,
    Route, Server,
    EndpointExt,
    tera::{TeraTemplating, TeraTemplate, Tera, Context}
};

#[handler]
fn hello(Path(name): Path<String>, tera: Tera) -> TeraTemplate {
    let mut context = Context::new();
    context.insert("name", &name);
    tera.render("index.html.tera", &context)
}

#[tokio::main]
async fn main() -> Result<(), std::io::Error> {
    let app = Route::new()
        .at("/hello/:name", get(hello))
        .with(TeraTemplating::from_glob("templates/**/*"));

    Server::new(TcpListener::bind("127.0.0.1:3000"))
        .run(app)
        .await
}
```

Christoph-AK:
[...]

PR #515, opened 2023-02-18
Feat/tera: Support Tera Templating

Galitan-dev:
# Feat/tera: Support Tera Templating
See [Issue #512](https://github.com/poem-web/poem/issues/512)
## Description
I added Tera Templating support (the Tera instance is sent to handlers via the request extensions). I also implemented the i18n tera filters as said in [my last PR](https://github.com/poem-web/poem/pull/507).
## Usage Example
(See the new [tera-i18n example](https://github.com/Galitan-dev/poem/tree/feat/tera/examples/poem/tera-i18n))
```yaml
# ressources/en-US/simple.ftl
hello-world = Hello world!
welcome = welcome { $name }!
```
```html
<!-- template/hello.html.tera ->
<h1>{{ "welcome" | translate(name=name) }}</h1>
```
```rust
// src/main.rs
#[handler]
fn hello(Path(name): Path<String>, tera: Tera) -> TeraTemplate {
    tera.render("hello.html.tera", &ctx!{ "name": &name })
}

#[tokio::main]
    let app = Route::new()
        .at("/hello/:name", get(hello))
        .with(TeraTemplating::from_glob("templates/**/*"))
        .using(filters::i18n::translate)
        .data(resources);

    Server::new(TcpListener::bind("127.0.0.1:3000"))
        .run(app)
        .await
}
```


Galitan-dev:
oops I broke the Locale extractor, I thought that my implementation of FromRequest for FromRequestSync worked

Galitan-dev:
I think we are good

poem/src/tera/middleware.rs
```diff
@@ -0,0 +1,135 @@
...
+    /// ```no_compile
+    /// use poem::tera::TeraTemplating;
+    ///
+    /// let templating = TeraTemplating::from_glob("templates/**/*");
+    /// ```
+    pub fn from_glob(glob: &str) -> Self {
+        let tera = match Tera::new(glob) {
+            Ok(t) => t,
+            Err(e) => {
+                println!("Parsing error(s): {e}");
```
attila-lin, line 28: better with `traing::error!` ?

Galitan-dev: I didn't know about it

Galitan-dev: It is not implemented for Tera :/

Galitan-dev:
I force-pushed because the commit is for the next PR.

poem/src/web/mod.rs
```diff
@@ -326,6 +328,16 @@ pub trait FromRequest<'a>: Sized {
...
+
+    /// Extract from request head and body synchronously.
+    fn from_request_sync(_req: &'a Request, _body: &mut RequestBody) -> Result<Self> {
+        panic!("Not implemented, please implement one of from_request and from_request_sync");
+    }
+
+    /// Extract from request head synchronously.
+    fn from_request_without_body_sync(req: &'a Request) -> Result<Self> {
+        Self::from_request_sync(req, &mut Default::default())
+    }
```
sunli829, line 340: I don't think it's necessary, what do you think?

```suggestion

    /// Extract from request head and body synchronously.
    fn from_request_sync(_req: &'a Request, _body: &mut RequestBody) -> Result<Self> {
        panic!("Not implemented, please implement one of from_request and from_request_sync");
    }

    /// Extract from request head synchronously.
    fn from_request_without_body_sync(req: &'a Request) -> Result<Self> {
        Self::from_request_sync(req, &mut Default::default())
    }
```

PR #517, opened 2023-02-27
Pattern matching in OpenAPI function args

stepantubanov:
This is my attempt at adding support for pattern matching in function arguments.

Example code:
```rust
    #[OpenApi]
    impl Api {
        #[oai(path = "/test3/:user_id", method = "get")]
        async fn test3(&self, Path(user_id): Path<i32>) {
            assert_eq!(user_id, 7);
        }
    }
```

In current master this code doesn't compile ("Invalid param definition"), I updated the attribute macro code to allow it.


sunli829:
thanks 🙂

PR #524, opened 2023-03-03
Allow not falling back to index file in StaticFilesEndpoint

Subserial:
The default for this option was true and could not be changed.


attila-lin:
Thanks!

PR #526, opened 2023-03-07
fix: alive connections counter for graceful shutdown

DDtKey:
It's currently looks a bit fragile:
It's possible to spawn new future, but parallel task will finish before spawned task is polled. 
So this check will notify waiter & stop server:
```rs
if alive_connections.fetch_sub(1, Ordering::SeqCst) == 1 {
    notify.notify_one();
}
```

It's hard to test, however sometimes we're encountering `connection reset` errors on client end, while server side almost immediately logged correct graceful shutdown & server stopped (we use large enough timeout, so it's not the case)

I believe it's more reliable & safe to increase counter before spawning the task of serving.

### DIFF
--- a/poem/src/server.rs
+++ b/poem/src/server.rs
@@ -130,7 +130,7 @@ where
                 },
                 res = acceptor.accept() => {
                     if let Ok((socket, local_addr, remote_addr, scheme)) = res {
-                        alive_connections.fetch_add(1, Ordering::SeqCst);
+                        alive_connections.fetch_add(1, Ordering::Release);
 
                         let ep = ep.clone();
                         let alive_connections = alive_connections.clone();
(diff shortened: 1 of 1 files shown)


poem/src/server.rs
```diff
@@ -135,9 +135,8 @@ where
                         let notify = notify.clone();
                         let timeout_notify = timeout_notify.clone();
 
+                        alive_connections.fetch_add(1, Ordering::SeqCst);
```
DDtKey, line 138: Btw, I see `SeqCst` is used here and it's strict. 
Probably it should be refactored to `Acquire/Release` 🤔

sunli829: Could you help me with this? 🙂

DDtKey: Sure!
As separate PR or this one?

sunli829: thanks 🙂

Issue #536, opened 2023-03-18
Error when using other Tower Service with Poem


banool:
See this code:
```
use anyhow::{Context as AnyhowContext, Result};
use aptos_api::context::Context;
use aptos_protos::api::v2::{
    api_v2_server::{ApiV2, ApiV2Server},
    GetAccountModuleRequest, GetAccountModuleResponse,
};
use poem::{endpoint::TowerCompatExt, IntoEndpoint, Route};
use std::sync::Arc;
use tonic::{transport::Server, Request, Response, Status};

#[derive(Clone)]
pub struct ApiV2Service {
    pub context: Arc<Context>,
}

pub fn build_api_v2_service(context: Arc<Context>) -> Result<impl IntoEndpoint> {
    let service = ApiV2Service { context };

    let tower_service = Server::builder()
        .add_service(ApiV2Server::new(service))
        .into_service();

    let tower_service = tower_service.compat();

    let routes = Route::new().nest("/", tower_service);
    Ok(routes)
}

#[tonic::async_trait]
impl ApiV2 for ApiV2Service {
    async fn get_account_module(
        &self,
        request: Request<GetAccountModuleRequest>,
    ) -> Result<Response<GetAccountModuleResponse>, Status> {
        unimplemented!();
    }
}
```

Here you can see I define a service using tonic, convert it into a tower service, and then apply `.compat()` to it to make it compatible with Poem. Unfortunately, this isn't working: https://gist.github.com/banool/52c92650a00b6ac2d7322e1220ac9588. It seems like Poem requires that the service be Sync, but the one produced by Tonic is not.

Is there any way I can make this work?

banool:
I can see a possible solution in the Tonic side too: https://github.com/hyperium/tonic/issues/1322.

banool:
Of course if we had https://github.com/poem-web/poem/issues/535, this could be solved the other way around, where instead of Poem at the top, we have something else like Axum or Tower directly, and then Poem and Tonic underneath it.

banool:
For context, this problem makes it that the only solution is to run two separate web servers and then have a reverse proxy (on the same machine) in front of them.

PR #537, opened 2023-03-19
Remove Sync bound on tower compat

banool:
Addresses https://github.com/poem-web/poem/issues/536.


banool:
Actually this doesn't work since Endpoint must be Sync.

PR #544, opened 2023-03-23
Support for ipnet crate + IpAddr

mzachar:
Added support for ipnet crate

Also implemented Type for IpAddr so both ipv4 or ipv6 address can be used in single field/parameter


sunli829:
Thanks! 🙂

PR #550, opened 2023-03-28
fix(server): graceful shutdown of established connections

### LINKED ISSUE #546: Graceful shutdown for established connections
## Expected Behavior

`Poem` should trigger hyper's [graceful showdown method for connection](https://docs.rs/hyper/latest/hyper/server/conn/struct.Connection.html#method.graceful_shutdown) as soon as signal is received.
It handles correct shutdown: `GOAWAY` for `http/2` or disabling `keep-alive` for `http/1`

There is an example in hyper - [link](https://github.com/hyperium/hyper/blob/95a153bbc2717bd4233486e09848622ceb900574/examples/upgrades.rs#L168) (not exhaustive though)

## Actual Behavior

Request can be sent over already established connection after initiating graceful shutdown.
Currently [connection is just awaited ](https://github.com/poem-web/poem/blob/10de9d11daceed3def451999ea8496b636f7d1b2/poem/src/server.rs#L196) without any logic for that.

Just reuse the established connection  for new request after graceful shutdown is initiated.

## Specifications

- Version: up to 1.35 (current main)
- Platform: -
- Subsystem: -

DDtKey:
Closes #546

### DIFF
--- a/poem/src/server.rs
+++ b/poem/src/server.rs
@@ -11,9 +11,9 @@ use http::uri::Scheme;
 use hyper::server::conn::Http;
 use tokio::{
     io::{AsyncRead, AsyncWrite, Result as IoResult},
-    sync::Notify,
     time::Duration,
 };
+use tokio_util::sync::CancellationToken;
 
 use crate::{
     listener::{Acceptor, AcceptorExt, Listener},
@@ -93,9 +93,9 @@ where
         let Server { listener, name } = self;
         let name = name.as_deref();
         let alive_connections = Arc::new(AtomicUsize::new(0));
-        let notify = Arc::new(Notify::new());
-        let timeout_notify = Arc::new(Notify::new());
-        let conn_shutdown_notify = Arc::new(Notify::new());
+        let cancellation_token = CancellationToken::new();
+        let timeout_token = CancellationToken::new();
+        let conn_shutdown_token = CancellationToken::new();
 
         let mut acceptor = match listener {
             Either::Listener(listener) => listener.into_acceptor().await?.boxed(),
@@ -112,7 +112,7 @@ where
         loop {
             tokio::select! {
                 _ = &mut signal => {
-                    conn_shutdown_notify.notify_waiters();
+                    conn_shutdown_token.cancel();
                     if let Some(timeout) = timeout {
                         tracing::info!(
                             name = name,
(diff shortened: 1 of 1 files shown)


poem/src/server.rs
```diff
@@ -93,8 +93,9 @@ where
         let Server { listener, name } = self;
         let name = name.as_deref();
         let alive_connections = Arc::new(AtomicUsize::new(0));
-        let notify = Arc::new(Notify::new());
-        let timeout_notify = Arc::new(Notify::new());
+        let cancellation_token = CancellationToken::new();
+        let timeout_token = CancellationToken::new();
+        let conn_shutdown_token = CancellationToken::new();
```
DDtKey, line 98: I refactored it to `CancellationToken` instead of `Notify` because:
- `Notify::notify_waiters()` will notify only registered `Notified` (i.e `notify.registred()` is already called)
- We spawn tokio task to serve connection and register it inside (like `timeout_notify.notified()`), while spawned task can be polled with delay

So it's just more reliable structure here

poem/src/server.rs
```diff
@@ -135,20 +138,25 @@ where
                         let ep = ep.clone();
                         let alive_connections = alive_connections.clone();
                         let notify = notify.clone();
-                        let timeout_notify = timeout_notify.clone();
+                        let timeout_token = timeout_token.clone();
+                        let conn_shutdown_token = conn_shutdown_token.clone();
```
DDtKey, line 142: I refactored it to `CancellationToken` instead of `Notify` because:
- `Notify::notify_waiters()` will notify only registered `Notified` (i.e `notify.registred()` is already called)
- We spawn tokio task to serve connection and register it inside (like `timeout_notify.notified()`), while spawned task can be polled with delay

So it's just more reliable structure here

poem/src/server.rs
```diff
@@ -135,20 +138,25 @@ where
...
                             } else {
-                                serve_connection(socket, local_addr, remote_addr, scheme, ep).await;
+                               serve_connection.await;
                             }
 
                             if alive_connections.fetch_sub(1, Ordering::Acquire) == 1 {
-                                notify.notify_one();
+                                // we have to notify only if there is a registered waiter on shutdown.
+                                // `notify_one` will make `notified` completed immidiately as soon as we don't have at least 1 connection
+                                notify.notify_waiters();
```
DDtKey, line 159: It was broken logic actually.

Once you have handled some number of connections - it will `notify_one` in advance. So gracefult shutdown didn't work as expected and just passed `notified()` immediately without waiting for notification

sunli829:
Thanks! 🙂

PR #551, opened 2023-04-01
Fixes:  the `Content-Type` requested by the client is not supported: multipart/form-data; boundary=--

mzachar:
Fixes `#[derive(Multipart)]` for struct, so it will work with `#[derive(ApiRequest)]`:

```
#[derive(ApiRequest)]
enum MyRequest {
    Json(Json<Data>),
    File(DataUpload),
}

#[derive(Multipart)]
struct DataUpload {
    file: Upload,
}
```
MyRequest will now correctly recognize content type `multipart/form-data; boudary=--....` as part of MyRequest::File


sunli829:
Thanks! 🙂

PR #555, opened 2023-04-06
chore: bump syn to 2.X

attila-lin:


attila-lin:
https://github.com/TedDriggs/darling/pull/226

PR #563, opened 2023-04-28
Add support for wildcard '*' in cors domains

Sagebati:
Hi ! 

This new feature is especially useful for applications that utilize multiple subdomains, as it eliminates the need to manually configure CORS for each subdomain. By simply adding the wildcard (*) to your CORS configuration, you can enable cross-domain requests for all subdomains of a given domain. I replaced the `HashSet::contains` search with a Regex containing all the domains (example:https://regex101.com/r/ZEhBy4/2).  I'm aware of `the allow_origin_fn`, but in opinion this is ergonomic.  The PR is almost ready to be merged only need a review of the `unwrap` used.

Any feedback is welcome.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,10 +20,10 @@ repository = "https://github.com/poem-web/poem"
 rust-version = "1.64"
 
 [workspace.dependencies]
-poem = { path = "poem", version = "1.3.55", default-features = false }
-poem-derive = { path = "poem-derive", version = "1.3.55" }
-poem-openapi-derive = { path = "poem-openapi-derive", version = "2.0.25" }
-poem-grpc-build = { path = "poem-grpc-build", version = "0.2.18" }
+poem = { path = "poem", version = "1.3.56", default-features = false }
+poem-derive = { path = "poem-derive", version = "1.3.56" }
+poem-openapi-derive = { path = "poem-openapi-derive", version = "2.0.27" }
+poem-grpc-build = { path = "poem-grpc-build", version = "0.2.20" }
 
 tokio = "1.17.0"
 serde_json = "1.0.68"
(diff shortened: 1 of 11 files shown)


attila-lin:
Hi, @Sagebati , Thank you for your mr.

Would you please add some testcase and fix the CI?

Sagebati:
Hi, I corrected the CI errors.

Sagebati:
Seems like the CI passes

DDtKey (commented):
Just a suggestion: consider to use a [wildmatch](https://crates.io/crates/wildmatch) crate

Sagebati:
I didn't knew this crate, I could rewrite the feature to use it. The change that it implies is we will have a `Vec<WildMatch>` instead of only one regex, but it easier to maintain. Let met know your preferences

sunli829:
What do you think if it could be better like below? @Sagebati 

```rust
pub struct Cors {
    allow_credentials: bool,
    allow_origins: HashSet<HeaderValue>,
    allow_origins_wildcard: Vec<WildMatch>, //////// <<<<<<<<<
    allow_origins_fn: Option<Arc<dyn Fn(&str) -> bool + Send + Sync>>,
    allow_headers: HashSet<HeaderName>,
    allow_methods: HashSet<Method>,
    expose_headers: HashSet<HeaderName>,
    max_age: i32,
}

impl Cors {
    fn allow_origin_to_regex(mut self, wildcard: impl AsRef<str>) -> Self {
        self.allow_origins_wildcard.push(WildMatch::new(wildcard));
        self
    }
}
```

sunli829:
Could you help to add a test case? 🙂 @Sagebati

sunli829:
Thanks! @Sagebati  🙂

PR #569, opened 2023-05-06
feat: upsert default_headers to dynamically change while testing

chikko80:
## Summary
This pull request adds a new function, upsert_default_header(), to the TestClient struct. This function allows users to update or insert a default header value for each request, making it easier to rotate the Authorization header with different access tokens during testing.

## Changes
Add upsert_default_header() function to TestClient struct
Include example usage of upsert_default_header() in comments

## Scenario
While testing endpoints with User Auth I want to be able to dynamically change the Auth Header to simulate different users with different roles/access tokens.

```rust
    pub async fn set_auth(&mut self, user_id: i32, auth_role: AuthRole) {
        let (access_token, _) =
            AuthController::create_login_creds(&self.database, &self.secret, user_id, auth_role)
                .await
                .unwrap();

        self.test_client
            .upsert_default_header("Authorization", format!("Bearer {}", access_token));
    }
```

Please review the changes and let me know if you have any feedback or concerns.

### DIFF
--- a/poem/src/test/client.rs
+++ b/poem/src/test/client.rs
@@ -71,7 +71,7 @@ impl<E: Endpoint> TestClient<E> {
         self
     }
 
-    /// Upserts the default header for each requests.
+    /// Upsert on default_headers for the current client.
     ///
     /// # Examples
     ///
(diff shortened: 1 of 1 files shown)


sunli829:
thanks! 🙂

PR #570, opened 2023-05-07
Trivial doc change to add `test` to the list of optional

navicore:
Trivial doc change to add `test` to the list of optional crate features.

I found helpful examples for usage of `poem::test::TestClient` when making my first poem endpoint tests but it wasn't obvious I needed to enable the `test` feature since it wasn't included in this list.


attila-lin:
Thank you~

PR #593, opened 2023-06-22
poem-grpc: request method should be post

gwik:
Request method for GRPC requests changed to POST.

As per spec:

https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

Some http proxies trim the body of a GET request it result in an error.
Also some servers might only accept a GET request.


sunli829:
Thanks!😁

gwik:
Thank you for this great framework! I appreciate your work.

PR #605, opened 2023-07-12
Allow ApiExtractor to provide SecurityScheme and Parameter simultaneously

amtelekom:
Hi,
we have a configuration Parameter that is dependent on both the Authorization header, as well as a parameter Value.
This PR implements an additional ApiExtractorType that allows injecting both a SecurityScheme as well as a Parameter into an OpenAPI-Spec simultanously.


sunli829:
Although not a good solution, there is no breaking change, and I will fix it in the next major version. 🙂

PR #611, opened 2023-07-27
(Trivial) fix example curl commands in the README.md

eudoxia0:
The URLs were wrong.

Additionally, I couldn't get `cargo run --example hello_world` to work in either the root or the `poem-openapi/` directory, but maybe that's my version of `cargo` or something.


sunli829:
> Additionally, I couldn't get cargo run --example hello_world to work in either the root or the poem-openapi/ directory, but maybe that's my version of cargo or something.

Because all the examples are in a separate workspace.

PR #625, opened 2023-08-08
Added ability to return an error for failed `SecuritySchema` checker.

NexRX:
Hi, I thought I'd give a crack at adding support for returning errors for the Security Schema checker.

I've never writen a proc macro before so let me know if there is a better way to implictly take a Option or Result without knowing what you're getting at compile time (_I really tried to find out :(_)

Thanks!


sunli829:
Looks great, can you add a test case to [here](https://github.com/poem-web/poem/blob/master/poem-openapi/tests/security_scheme.rs)? 🙂

NexRX:
Pushed those tests 🤖. Was waiting for feedback before commiting. Also updated security scheme markup doc ✨

sunli829:
good job, thank you! ♥

sunli829:
format code with `cargo +nightly fmt`

NexRX:
Is the pipeline failing because of me? Because running `cargo +nightly fmt` locally I can see 261 changes when i only chnaged 3 files. Ill push it anyway, let me know if you want it reset.

sunli829:
We should move CheckerReturn type to poem-openapi crate and expose it in poem::__private, otherwise there may be an error of duplicate type definition. 🙂

NexRX:
> ~We should move CheckerReturn type to poem-openapi crate and expose it in poem::__private, otherwise there may be an error of duplicate type definition.~
> 
> This problem should not exist, but in order to reduce compilation time, it is still hoped that the `CheckerReturn` type can be moved to `poem-openapi`. 🙂

It does feel silly having it there, I initially put it there just to how to implicitly get it working, but ill move that now.

NexRX:
Awsome! ✨

Issue #645, opened 2023-09-04
Issue about endpoint path suffix "/" 


RWDai:
## Different behavior with 2.0.X and 3.0.4

same code:
``` rust
use poem::{listener::TcpListener, Route, Server};
use poem_openapi::{payload::PlainText, OpenApi, OpenApiService};

struct Api;

#[OpenApi(prefix_path = "/hello")]
impl Api {
    /// Hello world
    #[oai(path = "/", method = "get")]
    async fn index(&self) -> PlainText<&'static str> {
        PlainText("Hello World")
    }
}


#[tokio::main]
async fn main() {
    let api_service =
        OpenApiService::new(Api, "Hello World", "1.0").server("http://localhost:3000");
    let ui = api_service.swagger_ui();
    let app = Route::new().nest("/", api_service).nest("/docs", ui);

    Server::new(TcpListener::bind("127.0.0.1:3000"))
        .run(app)
        .await;
}
```
### In 2.0.x
You can access the `hello` endpoint through "http://127.0.0.1:3000/hello", but "http://127.0.0.1:3000/hello/" does not work
### In 3.0.4
Just the opposite is that you can access through "http://127.0.0.1:3000/hello/", and "http://127.0.0.1:3000/hello" does not work

This will make the cost of code changes too high when upgrading the version.

### question
The path parameter in the oai macro above Api.index cannot be left blank and cannot be empty. So how to remove the / from the Api.index path suffix?

## Specifications

- Version: 3.0.4
- Platform: platform independent

sunli829:
fixed in `v3.0.5`

PR #646, opened 2023-09-04
Allow using empty path in openapi

4t145:
This pr is to slove the issue https://github.com/poem-web/poem/issues/645
May us allow the value of `path` parameter in `oai` macro to be empty since `/somepath` and  `/somepath/` are considered different?


sunli829:
thanks! 🙂

sunli829:
I fixed #645, it was a bug, so I closed this PR, still thank you. 🙂

Issue #660, opened 2023-09-18
Misplaced `</html>` in swagger_ui HTML template


SaculRennorb:
https://github.com/poem-web/poem/blob/51d6d8d934aecf093e8044d9ba78e05349473b4c/poem-openapi/src/ui/swagger_ui/mod.rs#L15C1-L15C8

has the closing html tag misplaced. i would open a pr, but its hardly worth the fork.

```diff
  const SWAGGER_UI_TEMPLATE: &str = r#"
  <html charset="UTF-8">
  <head>
      <meta http-equiv="Content-Type" content="text/html;charset=utf-8">
      <title>Swagger UI</title>
      <style charset="UTF-8">{:style}</style>
      <script charset="UTF-8">{:script}</script>
  </head>
- </html>
  <body>
  
  <div id="ui"></div>
  <script>
      let spec = {:spec};
      let oauth2RedirectUrl;
  
      let query = window.location.href.indexOf("?");
      if (query > 0) {
          oauth2RedirectUrl = window.location.href.substring(0, query);
      } else {
          oauth2RedirectUrl = window.location.href;
      }
  
      if (!oauth2RedirectUrl.endsWith("/")) {
          oauth2RedirectUrl += "/";
      }
      oauth2RedirectUrl += "oauth-receiver.html";
  
      SwaggerUIBundle({
          dom_id: '#ui',
          spec: spec,
          filter: false,
          oauth2RedirectUrl: oauth2RedirectUrl,
      })
  </script>
  
  </body>
+ </html>
  "#;
```

Also from what I gathered it seems like it might be possible to replace the html template for the UI as a consumer of the package, but I'm not quite certain if that is possible and if so how to accomplish it, so I might aswell ask:
 
Is is possible to provide a custom HTML template to set things like the page title and icon for those UIs (Specifically the swagger one in my case)?

sunli829:
You can open a PR and any fixes are welcome. 🙂

SaculRennorb:
I mean i put the fix in my original comment, but sure - i'll open a pr when i find some time in the next few days.

SaculRennorb:
I will close this as completed as the original problem is now fixed, but my second question still stands:

> Is is possible to provide a custom HTML template to set things like the page title and icon for those UIs (Specifically the swagger one in my case)?

PR #662, opened 2023-09-25
Addressing misplaced `<html>` tag in swagger template (#660)

SaculRennorb:
Addressing https://github.com/poem-web/poem/issues/660


attila-lin:
thank you

PR #666, opened 2023-10-05
count string length correctly in OpenAPI validators

mishmish-dev:
Currently in the poem's OpenAPI module, derived **maxLength** and **minLength** validators use `str::len()` method, which counts UTF-8 bytes, not Unicode characters in string. While **JSON Schema** spec defines **maxLength** and **minLength** for number of characters.

This pull request fixes that, also adding extra couple unit tests to better cover edge cases.

### DIFF
--- a/poem-openapi/src/validation/max_length.rs
+++ b/poem-openapi/src/validation/max_length.rs
@@ -21,7 +21,7 @@ impl MaxLength {
 impl<T: AsRef<str>> Validator<T> for MaxLength {
     #[inline]
     fn check(&self, value: &T) -> bool {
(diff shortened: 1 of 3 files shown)


mishmish-dev:
Hi! I want to leave an extra remark.

The proposed implementation assumes that **character** in Unicode string is the same as **Unicode scalar value**, which, I think, is pretty reasonable interpretation in  the context of web APIs.

However, I am not completely sure it is precisely JSON Schema / JSON specs' interpretation. (For example, is Indonesian flag emoji 🇮🇩 one character or two?)

sunli829:
thanks

PR #668, opened 2023-10-10
added permissions and owner to UnixListener

yasamoka:
This helps avoid having to set permissions and ownership on the socket after it is created.

Thank you!


sunli829:
thanks

Issue #664, opened 2023-10-01
In the examples, listen on 0.0.0.0 instead of 127.0.0.1


schitcrafter:
Listening from localhost (127.0.0.1) might work well for a development environment, but when deployed in a docker container the service will not be accessible from outside. Especially for people that know little about networking / docker in general, something like this can be very frustrating to debug - the container will be running, networking will be correctly setup, but every connection is refused.

attila-lin:
Nice catch. 

PR are welcome!

schitcrafter:
Alright, I'll make a PR - just to be sure, I didn't mention it in the original issue, listening on 0.0.0.0 means that the developer will most likely expose their poem service on the local network (in the case where it's not running in a container). Just wanted to mention that.

attila-lin:
Yes, we may add some comments on `0.0.0.0` to notify the newbie.

schitcrafter:
Not sure if I've forgotten anything, but this should be everything in examples/ plus README's and a couple doc comments

attila-lin:
seems finished.

Please reopen if anything is wrong.

PR #672, opened 2023-10-12
In the examples, listen on 0.0.0.0 instead of 127.0.0.1 #664

schitcrafter:
#664

### DIFF
--- a/examples/poem/mongodb/src/main.rs
+++ b/examples/poem/mongodb/src/main.rs
@@ -60,13 +60,13 @@ async fn main() -> io::Result<()> {
     }
     tracing_subscriber::fmt::init();
 
-    let mongodb = Client::with_uri_str("mongodb://127.0.0.1:27017")
+    let mongodb = Client::with_uri_str("mongodb://0.0.0.0:27017")
         .await
         .unwrap()
         .database("test");
     let collection = mongodb.collection::<Document>("user");
 
-    Server::new(TcpListener::bind("127.0.0.1:3000"))
+    Server::new(TcpListener::bind("0.0.0.0:3000"))
         .run(
             Route::new()
                 .at("/user", get(get_users).post(create_user))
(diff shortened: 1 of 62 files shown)


examples/poem/mongodb/src/main.rs
```diff
@@ -60,13 +60,13 @@ async fn main() -> io::Result<()> {
     }
     tracing_subscriber::fmt::init();
 
-    let mongodb = Client::with_uri_str("mongodb://127.0.0.1:27017")
+    let mongodb = Client::with_uri_str("mongodb://0.0.0.0:27017")
```
viell-dev, line 63: Connecting to 0.0.0.0 isn't strictly speaking valid. (But, in practice, it should be equivalent to connecting to localhost.)

examples/poem/redis-session/src/main.rs
```diff
@@ -20,13 +20,13 @@ async fn main() -> Result<(), std::io::Error> {
     }
     tracing_subscriber::fmt::init();
 
-    let client = Client::open("redis://127.0.0.1/").unwrap();
+    let client = Client::open("redis://0.0.0.0/").unwrap();
```
viell-dev, line 23: Connecting to 0.0.0.0 isn't strictly speaking valid. (But, in practice, it should be equivalent to connecting to localhost.)

examples/poem/websocket-chat/src/main.rs
```diff
@@ -42,7 +42,7 @@ fn index() -> Html<&'static str> {
             sendForm.hidden = false;
             msgsArea.hidden = false;
             msgInput.focus();
-            ws = new WebSocket("ws://127.0.0.1:3000/ws/" + nameInput.value);
+            ws = new WebSocket("ws://0.0.0.0:3000/ws/" + nameInput.value);
```
viell-dev, line 45: Connecting to 0.0.0.0 isn't strictly speaking valid. (But, in practice, it should be equivalent to connecting to localhost.)

viell-dev (changes requested):
Looks like you did a search-and-replace. You also replaced some outgoing IP addresses.

sunli829:
thanks 😂

PR #709, opened 2023-12-19
allow WebSocket casing for upgrade header

tomlinton:
The RFC [allows the use of](https://datatracker.ietf.org/doc/html/rfc6455#section-11.2) `WebSocket` which some frameworks follow, e.g. [autobahn](https://github.com/crossbario/autobahn-python/issues/1489#issuecomment-861391942).


sunli829:
Thanks! 🙂

PR #717, opened 2024-01-02
feat(openapi): added example value support for param/schema

chirag-droid:
## Example Code

image

```rs
struct Api;

fn hello_world() -> String { String::from("Hello World") }

#[OpenApi]
impl Api {
    #[oai(path = "/", method = "get")]
    async fn get(
        &self,
        #[oai(example = "hello_world")] name: Query<String>,
    ) -> PlainText(String) {
        PlainText(name.0)
    }
}
```

### DIFF
--- a/poem-openapi-derive/src/api.rs
+++ b/poem-openapi-derive/src/api.rs
@@ -7,7 +7,7 @@ use syn::{
 };
 
 use crate::{
-    common_args::{APIMethod, CodeSample, DefaultValue, ExternalDocument, ExtraHeader},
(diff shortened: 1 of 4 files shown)


chirag-droid:
I have also added test for it in `poem-openapi/tests/operation_param.rs`

attila-lin:
fmt failed~

chirag-droid:
let me fix it. there should be a git pre commit hook for this.

sunli829:
thanks! 🙂

PR #737, opened 2024-01-26
update cookie to 0.18, use cloned() method

getong:
cloned() method see
https://rust-lang.github.io/rust-clippy/master/index.html#/map_clone


attila-lin:
LGTM

PR #749, opened 2024-02-15
Adding `serialize_with` and `deserialize_with` attributes to struct fields

### LINKED ISSUE #585: Add support for serialize_with and deserialize_with
## Description of the feature

I'm currently evaluating a migration of an existing web project over to poem/poem-openapi and I was wondering if there is an equivalent to serialize_with/deserialize_with for poem_openapi::Object

I've got a project library that uses an external type in one of its fields [Geozero](https://github.com/georust/geozero)

## Code example (if possible)

```rust

pub mod MyDatabaseModels {
  use poem_openapi::Object;
  use serde::{Deserialize, Serialize};
  use serde_json;
  use geozero::wkb::Ewkb;

  #[derive(Debug, Serialize, Deserialize, Object)
  pub struct MyLargeTableA {
    pub id: uuid::Uuid,
    pub name: String,
    #[serde(
        serialize_with = "geojson_serialize",
        deserialize_with = "geojson_deserialize",
    )]
    pub geojson: Option<Ewkb>,
    // dozens of additional fields after
  };
}

```

I added #[derive(Object)] as above to MyLargeTableA but got four errors:

- the trait `Type` is not implemented for `geozero::wkb::Ewkb`
- variant or associated item cannot be called on `std::option::Option<geozero::wkb::Ewkb>` due to unsatisfied trait bounds
- the trait `ParseFromJSON` is not implemented for `geozero::wkb::Ewkb`
- the trait `ToJSON` is not implemented for `geozero::wkb::Ewkb`

I'm relatively new to rust unfortunately, but from what I understand the serde serialize_with/deserialize_with is what I'm looking for with #[oai]

Is there a better/cleaner solution?

SaculRennorb:
Heads up: I did not run all tests locally as the dependencies don't properly resolve without installing additional stuff, but the 'object' tests run fine (if you specify the `time` dependency).

I fully understand if you don't want to merge this in the public version, as it does not come with safety features and allows the consumer of the library to violate the generated OAI spec (see comment in tests).

Also I didn't figure out where to put the documentation, so if you point me to it I will still add that.

closes #585 as far as I can tell.

### DIFF
--- a/poem-openapi-derive/src/object.rs
+++ b/poem-openapi-derive/src/object.rs
@@ -42,6 +42,10 @@ struct ObjectField {
     skip_serializing_if_is_empty: bool,
     #[darling(default)]
     skip_serializing_if: Option<Path>,
+    #[darling(default)]
+    serialize_with: Option<Path>,
+    #[darling(default)]
+    deserialize_with: Option<Path>,
 }
 
 #[derive(FromDeriveInput)]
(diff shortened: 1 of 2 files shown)


sunli829:
thanks 🙂

PR #753, opened 2024-02-19
poem-openapi: Replace `naive-date` with `date`

sawi97:
When a date type was specified in an openapi Object, it was previously set to the format `naive-date`, which isn't recognized by doc-generators.

This replaces the openapi text format `naive-date` with `date`, aligning with the [OAS definition](https://spec.openapis.org/oas/v3.0.3#data-types) which uses RFC3339 full-date (which is naive by def).

It enables doc-generators (e.g. rapidoc) to generate examples when the schema contains a `time::Date` or `chrono::NaiveDate` field.

### DIFF
--- a/poem-openapi/src/types/external/chrono.rs
+++ b/poem-openapi/src/types/external/chrono.rs
@@ -143,7 +143,7 @@ macro_rules! impl_naive_datetime_types {
 }
 
 impl_naive_datetime_types!(NaiveDateTime, "string", "naive-date-time", "{:?}");
-impl_naive_datetime_types!(NaiveDate, "string", "naive-date", "{}");
+impl_naive_datetime_types!(NaiveDate, "string", "date", "{}");
 impl_naive_datetime_types!(NaiveTime, "string", "naive-time", "{}");
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


sawi97:
I believe openapi uses the same date/time definitions as [json schema](https://json-schema.org/understanding-json-schema/reference/string#dates-and-times). 

A naive date-time does not exist (yet). And I do not think it would be a good idea to set it to `date-time`, which is specified to contain timezone. Poem would not parse a `date-time` with time zone as a naive `date-time`, so the doc-generated examples would fail. And if there are any external input/output validators, they will most probably cause the request to fail.
The same goes for `time` (even tough time is not in the OAS specification, but rapidoc uses it).

So in my opinion the other date/time formats are fine, at least until OAS/JSON Schema has a specification for naive formats.

attila-lin:
Thanks for your reply!

sunli829:
If you expect a datetime in `RFC3339` format, should use `OffsetDateTime` instead of `PrimitiveDateTime`.

sawi97:
@sunli829 this PR is about date, not datetime. 
All `date` objects are naive in OAS and that is what this PR fixes. 

Can you reopen and consider merging?

GunnarMorrigan:
@sunli829, @sawi97 is right. 

Currently poem openapi translates a date object to string($naive-date) which is nothing in terms of Swagger.
Only a date-time and date object exist.

A time::Date, chrono::NaiveDate should be translated to string($date) for it to be inline with the swagger specifications.

PR #754, opened 2024-02-20
Add example of starting a local server and opening a browser

ilyagr:
Additionally, this demonstrates how to pick an open port, since this is often relevent for local servers (Jupyter-like applications).

Cc: https://github.com/poem-web/poem/discussions/751


attila-lin:
Thanks

Issue #762, opened 2024-02-28
May enter an infinite loop when parsing tls key.


4t145:
## Expected Behavior
Return an error.

## Actual Behavior
If the tls key beginning with an `Item::X509Certificate`, it may enter an infinite loop.

On line 77 of the code below, every loop it creates a new reader, so if the first item is a certification, it will enter an infinite loop.

https://github.com/poem-web/poem/blob/e3b2201f9c7ce4d9e59d69d653471ec2cf551ee1/poem/src/listener/rustls.rs#L76-L89

I think it should be something like this:
```rust
let mut key_reader = self.key.as_slice();
let priv_key = loop { 
     match rustls_pemfile::read_one(&mut key_reader)? { 
         Some(Item::Pkcs1Key(key)) => break key.into(), 
         Some(Item::Pkcs8Key(key)) => break key.into(), 
         Some(Item::Sec1Key(key)) => break key.into(), 
         None => { 
             return Err(IoError::new( 
                 ErrorKind::Other, 
                 "failed to parse tls private keys", 
             )) 
         } 
         _ => continue, 
     } 
 }; 
```
## Steps to Reproduce the Problem

1. use an x509 certification as a tls key

## Specifications

- Version: v2.0.0
- Platform: ubuntu 22.04 on wsl
- Subsystem: wsl

PR #763, opened 2024-02-28
Fix entering infinite loop when parsing a wrong tls key (#762).

4t145:
move tle key reader out from loop to avoid infinity loop
https://github.com/poem-web/poem/issues/762


attila-lin:
Thanks

PR #764, opened 2024-02-29
Fixup to e3b2201f: prevent open::that from blocking, fix a typo, minor code simplification

ilyagr:
`open::that` can block, for example when `firefox` is the default browser
on Linux and no instance of `firefox` is running.

-------------
Perhaps `tokio::spawn_blocking` with `open::that` would be better in terms of error handling than using `that_detached`. The latter does not report an error if the browser exits with an error code. OTOH, with `open::that`, killing the server might kill the browser...


attila-lin:
Thanks

PR #773, opened 2024-03-17
feat(afit): replace `async-trait` with AFIT

TD-Sky:
An attempt to migrate poem to AFIT.

The current obstacle is there are some type errors come from trait object of `Future` and I don't know how to resolve it at present.

@sunli829 encouraged me to create this draft PR.

---

About the macro `trait-make`: it's a fork of [trait-variant](https://github.com/rust-lang/impl-trait-utils/tree/main/trait-variant) which can convert `async fn` in trait to `fn -> impl Future + Send` in place.


sunli829:
released in `poem 3.0.0`

Issue #661, opened 2023-09-19
graceful shutdown hang


sundy-li:
## Actual Behavior

```
2023-09-19T07:59:08.341593Z  INFO poem::server: poem-1.3.57/src/server.rs:140 initiate graceful shutdown name="http handler Query" timeout_in_seconds=1.0
2023-09-19T07:59:08.342161Z  INFO poem::server: poem-1.3.57/src/server.rs:190 wait for all connections to close. name="http h
andler Query"
```

## Steps to Reproduce the Problem

Let's look at the codes:

https://github.com/poem-web/poem/blob/51d6d8d934aecf093e8044d9ba78e05349473b4c/poem/src/server.rs#L135-L192

If a `signal ` is received, the loop will break.

But if the `alive_connections ` is not zero, we will wait for `notify` but `notify.notify_waiters();` will not be called anymore.

## Specifications

- Version: latest
- Platform:
- Subsystem:

zkldi:
Seconding this, this is extremely frustrating as it also ignores any timeout functionality.

DDtKey:
The code attached doesn't look like a reason.
Yes, once `signal` is received, the loop will break.

But any accepted connection is spawned in separate task, so this part will be called anyway:
```rust
if alive_connections.fetch_sub(1, Ordering::Acquire) == 1 {
    // We have to notify only if there is a registered waiter on shutdown
    notify.notify_waiters();
}
```
And new connections won't be accepted, since loop is stopped. It also uses `timeout_token` - so once signal is received, all spawned connections will be dropped within the passed `timeout` to `run_with_graceful_shutdown`.

Do you use the timeout? Because otherwise connection may be just alive

DDtKey:
However I see possible race condition 🤔 
- `notify_waiters()` will notify only already registered waiters

Thus if it's called after `alive_connections.load(Ordering::Acquire) > 0`, but before `notify.notified()` - it might be never notified. 

But we also can't use `notify_one` which stores state that it's notified. Because `alive_connections` may reach `0` before shutdown and go up again (i.e there is no connection, but server is alive). It was [described here](https://github.com/poem-web/poem/pull/550#discussion_r1153494813)

PR #780, opened 2024-03-25
fix: possible race condition during graceful shutdown

DDtKey:
This PR just makes the behavior more reliable and kinda simplified, also addresses #661 (was able to repro manually and couldn't after the fix)

The possible race condition:
- `notify_waiters()` will notify only already registered waiters

Thus, if it's called after
```rust
drop(acceptor);
if alive_connections.load(Ordering::Acquire) > 0 {
```
But before `notify.notified()` - it might be never notified.

### DIFF
--- a/poem/src/server.rs
+++ b/poem/src/server.rs
@@ -166,7 +166,7 @@ where
                         let server_graceful_shutdown_token = server_graceful_shutdown_token.clone();
 
                         tokio::spawn(async move {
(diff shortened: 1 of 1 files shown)


poem/src/server.rs
```diff
@@ -178,8 +178,11 @@ where
                             }
 
                             if alive_connections.fetch_sub(1, Ordering::Acquire) == 1 {
```
dmitryvk, line 180: Shouldn't the ordering be `AcqRel`?

attila-lin:
I'd love to merge this. What's your opinion? @sunli829

PR #783, opened 2024-03-26
update MSRV to 1.75.0 in Cargo.toml

miikka:
Follow-up to ddfa81ea8323f0886a6f7212955d7504d4968083. Ran into this when trying to run tests on `master`.


attila-lin:
Thanks

PR #806, opened 2024-04-26
RequestId Middleware

vbrandl:
This PR introduces a feature-gated `RequestId` middleware that associates an unique ID with each incoming request. `ReqId` can be extracted in request handlers and returned as a response (e.g. for error pages). Each response contains a header with the ID (default name: `x-request-id` but the header name is configurable).

The `Tracing` middleware was made aware of the request ID and adds it to the request span (for this to work, the order of the middlewares is important. `Tracing` must be applied _before_ `RequestId`.

If `ReuseId::Use` is set, the middleware will propagate the request ID from an incoming request. If the request contained an `x-request-id` header (or whatever header name is configured), this value will be used.

### DIFF
--- a/examples/poem/requestid/Cargo.toml
+++ b/examples/poem/requestid/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "example-requestid"
+version.workspace = true
+edition.workspace = true
+publish.workspace = true
+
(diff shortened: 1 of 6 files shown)


sunli829:
thanks 🙂

PR #808, opened 2024-05-01
Fix error response builder, missing `Content-Type` header

Mr-Leshiy:
Fixed an issue of missing `Content-Type` header for the `ResponseError` types:
- Fixed a default implementation of the `ResponseError::as_response` fn
- Fixed an implementation of the  `StaticFileError::as_response` fn


sunli829:
thanks 🙂

PR #816, opened 2024-05-13
Expose Poem-OpenApi Upload File struct

Nahuel-M:
Some crates, including the (Amazon S3 Crate)[https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/], expect the `File` struct directly, instead of the `impl AsyncRead`. This PR exposes this `File` struct through a getter function.

### DIFF
--- a/poem-openapi/src/types/multipart/upload.rs
+++ b/poem-openapi/src/types/multipart/upload.rs
@@ -79,7 +79,7 @@ impl Upload {
     }
 
     /// Consumes this body object to return the file.
-    pub fn get_file(self) -> File {
+    pub fn into_file(self) -> File {
         self.file
     }
 }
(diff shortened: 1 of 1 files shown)


poem-openapi/src/types/multipart/upload.rs
```diff
@@ -77,6 +77,11 @@ impl Upload {
     pub fn into_async_read(self) -> impl AsyncRead + Unpin + Send + 'static {
         self.file
     }
+
+    /// Consumes this body object to return the file.
+    pub fn get_file(self) -> File {
```
antoninfin, line 82: Since it consumes self, I think `into_file` is more idiomatic.

gwik, line 82: Since it consumes self, `into_file` is more idiomatic.

sunli829:
thanks!

PR #817, opened 2024-05-15
poem-grpc: message can span multiple frame

gwik:
When a response was big enough to be split into multiple frames, the response decoder was incorrectly checking the frame decoder for an incomplete frame before receiving the remaining bytes.


sunli829:
thanks! 🙂

PR #825, opened 2024-05-30
Fix `EmbeddedFilesEndpoint` not working for `index.html` in subdirectories

### LINKED ISSUE #824: EmbeddedFilesEndpoint not working for index.html in subdirectory
I have the following two files at the same level of Cargo.toml:
```
static/index.html
static/test/index.html
```

Then I wrote the following code:
```
use poem::{endpoint::EmbeddedFilesEndpoint, listener::TcpListener, Route, Server};
use rust_embed::RustEmbed;

#[derive(RustEmbed)]
#[folder = "static"]
pub struct Static;

#[tokio::main]
    if std::env::var_os("RUST_LOG").is_none() {
        std::env::set_var("RUST_LOG", "poem=debug");
    }
    tracing_subscriber::fmt::init();

    let app = Route::new().nest("/static", EmbeddedFilesEndpoint::<Static>::new());
    Server::new(TcpListener::bind("0.0.0.0:3000"))
        .run(app)
        .await
}

```

When I visit `http://127.0.0.1:3000/static`, the program can get `static/index.html` correctly.
When I visit `http://127.0.0.1:3000/static/test/index.html`, the program can also get `static/test/index.html` correctly.
But when I visit `http://127.0.0.1:3000/static/test` and `http://127.0.0.1:3000/static/test/`, I get 404 errors. But they should be correctly redirected to `static/test/index.html` just like `static/index.html`.
I think this is a bug.

czy-29:
Close #824 

----------------
This is a simple fix of this bug, which is more suitable as a default behavior of `EmbeddedFilesEndpoint`.

Its basic logic is:
First: `let path = req.uri().path().trim_start_matches('/');`
Then if `path` is empty, and `req.original_uri().path()` does not end with '/', add '/' to the original uri path and redirect.
Otherwise, if `req.original_uri().path()` ends with '/', add `index.html` to `path` and call `EmbeddedFileEndpoint`.
Otherwise check if the file pointed to `path` exists, and if so, call `EmbeddedFileEndpoint` directly.
Otherwise, add `/index.html` and check if the file exists. If it does, add '/' to the original uri path and redirect.
Otherwise, call the `EmbeddedFileEndpoint` to return an appropriate 404.

However, in the long run, [EmbeddedFilesEndpoint](https://docs.rs/poem/latest/poem/endpoint/struct.EmbeddedFilesEndpoint.html) should have the same interface and behavior as [StaticFilesEndpoint](https://docs.rs/poem/latest/poem/endpoint/struct.StaticFilesEndpoint.html) to ensure the consistency of the framework API, while also providing more customization of static resource behavior.

This fix only provides a more reasonable and consistent default behavior without introducing an API Breaking Change.

----------------
这是对这个bug的一个简易修复，比较适合作为`EmbeddedFilesEndpoint`的一个默认行为。

其基本逻辑是：
首先：`let path = req.uri().path().trim_start_matches('/');`
然后如果`path`为空，且`req.original_uri().path()`不以'/'结尾，则给original uri path加上'/'之后重定向。
否则如果`req.original_uri().path()`以'/'结尾，则给`path`加`index.html`然后调用`EmbeddedFileEndpoint`。
否则检查`path`指向的文件是否存在，如果存在则直接调用`EmbeddedFileEndpoint`。
否则加上`/index.html`之后再看看文件是否存在，存在则给original uri path加上'/'之后重定向。
否则直接调用`EmbeddedFileEndpoint`以返回合适的404。

但是长期来看，[EmbeddedFilesEndpoint](https://docs.rs/poem/latest/poem/endpoint/struct.EmbeddedFilesEndpoint.html)应当具备与[StaticFilesEndpoint](https://docs.rs/poem/latest/poem/endpoint/struct.StaticFilesEndpoint.html)相同的接口与行为，以保证框架API的一致性，同时也会提供对静态资源行为的更多定制性。

本修复只是在不引入API Breaking Change的前提下，提供了一个更合理且更一致的默认行为。

### DIFF
--- a/poem/src/endpoint/embed.rs
+++ b/poem/src/endpoint/embed.rs
@@ -83,16 +83,29 @@ impl<E: RustEmbed + Send + Sync> Endpoint for EmbeddedFilesEndpoint<E> {
     type Output = Response;
 
     async fn call(&self, req: Request) -> Result<Self::Output, Error> {
-        let mut path = req
-            .uri()
-            .path()
-            .trim_start_matches('/')
-            .trim_end_matches('/')
-            .to_string();
-        if path.is_empty() {
-            path = "index.html".to_string();
+        let path = req.uri().path().trim_start_matches('/');
+        let original_path = req.original_uri().path();
+        let original_end_with_slash = original_path.ends_with('/');
+
+        use header::LOCATION;
+
+        if path.is_empty() && !original_end_with_slash {
+            Ok(Response::builder()
+                .status(StatusCode::FOUND)
+                .header(LOCATION, format!("{}/", original_path))
+                .finish())
+        } else if original_end_with_slash {
+            let path = format!("{}index.html", path);
+            EmbeddedFileEndpoint::<E>::new(&path).call(req).await
+        } else if E::get(path).is_some() {
+            EmbeddedFileEndpoint::<E>::new(path).call(req).await
+        } else if E::get(&format!("{}/index.html", path)).is_some() {
+            Ok(Response::builder()
+                .status(StatusCode::FOUND)
+                .header(LOCATION, format!("{}/", original_path))
+                .finish())
+        } else {
+            EmbeddedFileEndpoint::<E>::new(path).call(req).await
         }
-        let path = path.as_ref();
-        EmbeddedFileEndpoint::<E>::new(path).call(req).await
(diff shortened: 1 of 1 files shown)


sunli829:
thanks

PR #839, opened 2024-06-30
fix(prometheus): use correct `content-type` header

SimenB:
The header should include the version.

https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format

https://docs.rs/prometheus/0.13.4/src/prometheus/encoder/text.rs.html#12-13


attila-lin:
Would you mind add some test case?

SimenB:
added 👍

SimenB:
hah, didn't notice the formatting thing. fix pushed

attila-lin:
Thanks！

PR #844, opened 2024-07-09
fix: catch unwind for serve_connection

### LINKED ISSUE #843: serve_connection should handle panic?
https://github.com/poem-web/poem/blob/c66d78a54812709768f4723f0bda0cc79a6b7c82/poem/src/server.rs#L168-L197

We have a situation that `serve_connection.await` panic and thus `alive_connections.fetch_sub` doesn't execute, so that outer `notify.notified().await` blocks forever.

We should either handle panic, or do a timeout for waiting connections to close.

tisonkun:
This closes #843.

### DIFF
--- a/poem/src/server.rs
+++ b/poem/src/server.rs
@@ -3,6 +3,7 @@ use std::{
     future::Future,
     io,
     io::IoSlice,
+    panic::AssertUnwindSafe,
     pin::Pin,
     sync::{
         atomic::{AtomicUsize, Ordering},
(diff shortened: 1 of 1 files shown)


tisonkun:
Tested locally. Work well.

attila-lin:
Can this condition add a test case?

tisonkun:
I don't know your test suites and conventions, so if you can add it or point the position of the cases, it would be helpful.

tisonkun:
And you can trigger the CI so that we know if it introduces some regressions.

attila-lin:
Looks fine. Thanks~

tisonkun:
Thank you! Perhaps you can trigger a release so that I'd upgrade to the new version and run the regression tests in my project also :D

Issue #845, opened 2024-07-13
Drop `#[derive(Default)]` from handler macro


crescentrose:
## Description of the feature

The `#[handler]` macro currently [adds a `Debug` derive](https://github.com/poem-web/poem/blob/master/poem-derive/src/lib.rs#L78) to every endpoint that it generates. This causes some issues when working with handlers that take generic arguments.

In some situations, e.g. services with shared database connections, a `Default` implementation is not appropriate or even possible. Based on some cursory testing, it seems like the `default()` method doesn't even get called. But it is very much possible that I missed something, so please let me know if a `Default` implementation is necessary.

To avoid a breaking change, perhaps the `Default` derive could only be omitted with an optional parameter to the `handler` macro? 

## Code example (if possible)

```rust
use poem::{get, handler, middleware::AddData, test::TestClient, web::Data, EndpointExt, Route};


trait MyTrait: Clone + Send + Sync + 'static {
    fn do_something(&self) -> String;
}

#[derive(Clone)]
struct MyService;

impl MyTrait for MyService {
    fn do_something(&self) -> String {
        String::from("hello!")
    }
}

impl Default for MyService { // 👈 I want to get rid of this
    fn default() -> Self {   //    since it does not seem to get called
        unreachable!()       //    and having this `unreachable!` here
    }                        //    makes me slightly uncomfortable.
}

#[handler]
async fn hello<T: MyTrait>(Data(service): Data<&T>) -> String {
    service.do_something()
}

#[tokio::main]
async fn main() {
    let tc = TestClient::new(
        Route::new()
            .at("/", get(hello::<MyService>::default()))
            .with(AddData::new(MyService)),
    );

    tc.get("/").send().await.assert_text("hello!").await;
}
```

thinety:
The `Default` implementation is needed to create the handler (`hello::<MyService>::default()`), but using [#[derive(Default)]` is not ideal](https://github.com/rust-lang/rust/issues/26925).

https://github.com/poem-web/poem/pull/848 fixes this.

PR #848, opened 2024-07-17
Add manual `Default` implementation for `#[handler]`

thinety:
Sometimes, [`#[derive(Default)]` uses incorrect/conservative bounds](https://github.com/rust-lang/rust/issues/26925). This PR adds a manual implementation of `Default` to the `#[handler]` proc macro.

For a concrete example of this restriction, see https://github.com/poem-web/poem/issues/845.

Fixes https://github.com/poem-web/poem/issues/845.


attila-lin:
Thank you!

Issue #801, opened 2024-04-13
Add CookieConfig to Csrf middleware


fanie42:
Can we add CookieConfig to the Csrf middleware too? Mainly I would like to be able to change the csrf token cookie name. I don't know how applicable all of the fields are, but maybe it can just be copied from Session middleware?

```rust
let app = Route::new()
        .nest("/", api_service)
        .with(ServerSession::new(CookieConfig::new().name("my_session_id_cookie_name").secure(false), MemoryStorage::new()))
        .with(Csrf::new(CookieConfig::new().name("my_csrf_token_cookie_name").secure(false)));
```

PR #864, opened 2024-08-07
feat: allow to set a custom name for the csrf cookie (#801)

brussee:
Added feature:
Allow to set a custom name for the csrf cookie.

Related to: #801

### DIFF
--- a/poem/src/middleware/csrf.rs
+++ b/poem/src/middleware/csrf.rs
@@ -104,6 +104,15 @@ impl Csrf {
         Default::default()
     }
 
+    /// Sets the name of the csrf cookie. Default is `poem-csrf-token`.
+    #[must_use]
+    pub fn cookie_name(self, value: impl Into<String>) -> Self {
+        Self {
+            cookie_name: value.into(),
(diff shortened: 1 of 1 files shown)


brussee:
Intended to be used for: https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#using-cookies-with-host-prefixes-to-identify-origins

poem/src/middleware/csrf.rs
```diff
@@ -104,6 +104,12 @@ impl Csrf {
         Default::default()
     }
 
+    /// Sets the name of the csrf cookie. Default is `poem-csrf-token`.
+    #[must_use]
+    pub fn cookie_name(self, cookie_name: String) -> Self {
```
attila-lin, line 109: I'd love receive something like `impl Into<String>` for better ergonomics

brussee: Done ✅

attila-lin:
Thank you!

Issue #701, opened 2023-11-30
Support for nullable data types (Option<>) in poem-openapi


ettom:
## Description of the feature
It would be nice if a type being wrapped in `Option<>` would be shown as nullable in the OpenAPI document. See https://stackoverflow.com/questions/48111459/how-to-define-a-property-that-can-be-string-or-null-in-openapi-swagger

## Code example
```rust
#[oai(path = "/hello", method = "get")]
async fn hello(&self) -> Json<Option<String>> {
    Json(None)
}
```
would generate

```yaml
paths:
  /hello:
    get:
      responses:
        '200':
          description: ''
          content:
            application/json; charset=utf-8:
              schema:
                type: string
                nullable: true
```

Additionally,
```rust
#[derive(Object)]
struct Foo {
    x: Option<i32>,
}

#[OpenApi]
impl Api {
    #[oai(path = "/hello", method = "post")]
    async fn hello(&self, _p: Json<Foo>) {}
}

```

would generate

```yaml
paths:
  /hello:
    post:
      requestBody:
        content:
          application/json; charset=utf-8:
            schema:
              $ref: '#/components/schemas/Foo'
        required: true
      responses:
        '200':
          description: ''
components:
  schemas:
    Foo:
      type: object
      properties:
        x:
          type: integer
          format: int32
          nullable: true
```

Thanks for the amazing work!

moritzruth:
By default, `serde` serializes `Option::None` as `null`. It does *not* skip the field altogether.

This means that objects serialized by `serde` actually do not adhere to the schema generated by `poem-openapi` in case there are `Option::None` values.

kidqueb:
> By default, serde serializes Option::None as null. It does not skip the field altogether.

Isn't this what is happening in the example and PR? A nullable field, albeit manually specified in the PR, would make `None` become `null`

PR #865, opened 2024-08-15
Implemented nullable fields for openapi spec generation

skytz:
I've added the ability to set nullable fields on a struct with poem_api::Object derive.

Following the spec here: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#fixed-fields-20

Either for each individual field, with `nullable` attibute, or at the struct level with `nullable_all`.

It also solves this opened issue: https://github.com/poem-web/poem/issues/701

### DIFF
--- a/poem-openapi-derive/src/object.rs
+++ b/poem-openapi-derive/src/object.rs
@@ -31,6 +31,8 @@ struct ObjectField {
     #[darling(default)]
     write_only: bool,
     #[darling(default)]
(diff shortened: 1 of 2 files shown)


kidqueb:
Gave this a shot and it worked great, thank you. Hopefully get this gets merged in soon.

sunli829:
thanks

PR #875, opened 2024-08-31
Grammatical fixes in middleware docstrings

bitfl0wer:
I have noticed some grammatical errors while reading through the documentation of this crate.
This PR is intended to fix the ones I noticed while not changing what they are supposed to convey.


sunli829:
thanks 🙂

PR #877, opened 2024-09-04
chore: bump deps and fix fmt

attila-lin:


jayvdb:
@attila-lin there are conflicts in this PR now

PR #878, opened 2024-09-05
fix: setting path pattern in opentelemetry traces

CompuIves:
In the OpenTelemetry Tracing middleware, the path pattern is not set as an attribute. I think that's because the `PathPattern` is initialized when the request is "called". Similar to [this PR](https://github.com/poem-web/poem/pull/462), I have moved the setting of the attribute to after the request is called.

As a second change (which might be considered breaking), I am renaming the span to `:method :pattern` instead of `:method :uri`. This is recommended by OpenTelemetry: https://opentelemetry.io/docs/specs/otel/trace/api/#span.

### DIFF
--- a/poem/src/middleware/opentelemetry_tracing.rs
+++ b/poem/src/middleware/opentelemetry_tracing.rs
@@ -96,14 +96,10 @@ where
             format!("{:?}", req.version()),
         ));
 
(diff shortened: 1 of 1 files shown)


sunli829:
thanks 🙂

PR #879, opened 2024-09-09
Fix typos

jayvdb:


.typos.toml
attila-lin, line 1: would you mind remove this file

jayvdb: This is the config for https://crates.io/crates/typos-cli

I can add CI for it if you like, or happy to remove if you are not interested in this

jayvdb: @attila-lin , could you confirm you really dont want this tool to work in this repo ?

jayvdb: ok, typos-cli config removed.

PR #886, opened 2024-09-17
Assign the description to the request object in OpenAPI

bruwozniak:
This fixes an issue where the doc comment description of an e.g. JSON parameter of a request would not show up in the generated API spec.

For example:
```rust
#[oai(path = "/users", method = "post", operation_id = "createUser")]
async fn create_user(
    &self,
    /// Parameters required to create a new user
    Json(params): Json<CreateUserParams>,
)
```
should translate to:
```yaml
  /users:
    post:
      requestBody:
        description: Parameters required to create a new user
        content:
          application/json; charset=utf-8:
            schema:
              $ref: '#/components/schemas/CreateUserParams'
        required: true
```
but without this change `description` is missing.

### DIFF
--- a/poem-openapi-derive/src/api.rs
+++ b/poem-openapi-derive/src/api.rs
@@ -396,9 +396,13 @@ fn generate_operation(
         });
 
         // request object meta
+        let param_desc = optional_literal(&param_description);
         request_meta.push(quote! {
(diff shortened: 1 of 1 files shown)


bruwozniak:
@sunli829 could you please take a look? it's a relatively small one

sunli829:
thanks

PR #909, opened 2024-11-01
Fix grammar in "a OpenAPI"

jakob753951:
Simple find+replace from
`a OpenAPI`
to
`an OpenAPI`

Also caught a place where the original capitalisation was `OpenApi`


sunli829:
thanks

PR #924, opened 2024-12-10
feat: Implement enable_type_name config method

mikhailzagurskiy:
Sometimes for infrastructure purposes we need to get name of the type to build topics or subjects in message queue, based on the type name. The `prost_build::Config` type has an ability to auto implement `prost::Name` trait for any type. To do so, the config has `enable_type_name()` method, but that method wasn't passed through `prost_grpc_build::Config`. I've added such method to the config and have added example to show the value


attila-lin:
Thank you!

PR #934, opened 2024-12-17
allow custom *Authorization types be used in SecurityScheme macro

ihor-rud:
Example usage:

```rust
struct Bar;

impl BearerAuthorization for Bar {
    fn from_request(req: &Request) -> poem::Result<Self> {
        todo!()
    }
}

#[derive(SecurityScheme)]
#[oai(ty = "bearer")]
struct Foo(Bar);
```


ihor-rud:
I am closing it because it looks like the same functionality can be achieved using `checker` parameter.

PR #936, opened 2024-12-22
Implement middleware for &T: Middleware<_>

horazont:
That allows to pass middlewares with their configuration around at application setup time without having to clone them.


attila-lin:
thank you

PR #940, opened 2024-12-23
Allows passing the style of a parameter in the openapi spec.

Nahuel-M:
Allows setting the desired parameter style as documented in the [swagger documentation](https://swagger.io/docs/specification/v3_0/serialization/).
Example:
``` rust
#[oai(path = "/hello", method = "get")]
async fn index(
    &self,
    #[oai(style = "deep_object")] Query(input): Query<String>,
) -> PlainText<String> {
    PlainText(format!("hello, world! {input}"))
}
```
This currently enables custom `ApiExtractor`s to extract several types of objects from the path, query, or headers. 
In the future, I hope to improve the default Query, Path and Header extractors to take the style into account.

### DIFF
--- a/poem-openapi-derive/src/api.rs
+++ b/poem-openapi-derive/src/api.rs
@@ -6,12 +6,12 @@ use syn::{
     Path, ReturnType, Type,
 };
 
-use crate::parameter_style::ParameterStyle;
 use crate::{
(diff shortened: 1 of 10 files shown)


attila-lin:
thank you

PR #943, opened 2025-01-03
Respect client cookie precedence

horazont:
The inline comment should explain everything, hopefully.


sunli829:
thank you so much! 🙂

PR #944, opened 2025-01-03
Set a Path on the CSRF cookie

horazont:
Without setting a Path, the UA will infer the Path from the URL path, meaning that it will take the "directory name" of a URL path (so for a response for a resource at `/index`, the path would be inferred as `/`, and for a response for `/foo/bar` it would be inferred to be `/foo/` etc.).

This is problematic with CSRF for multiple reasons:

- It causes a proliferation of CSRF cookies which pollute client-side storage.
- It breaks CSRF when requests are sent across paths. For example, if a resource at `/foo/bar` contains a form which submits to `/index`, they would be theoretically using different CSRF states.
- Poem only processes a single cookie, which means that we have to rely on the ordering specified in RFC 6265 Section 5.4. This is bad for two reasons:

  1. The ordering is only a SHOULD, not a MUST.
  2. Poem does, in fact, not process cookies in the correct ordering (this is subject of another commit to fix).

Note that this commit may break applications if they share a CSRF cookie name with other applications hosted at different paths on the same domain.

### DIFF
--- a/poem/src/middleware/csrf.rs
+++ b/poem/src/middleware/csrf.rs
@@ -1,4 +1,4 @@
-use std::{sync::Arc, time::Duration};
+use std::{borrow::Cow, sync::Arc, time::Duration};
 
 use base64::{engine::general_purpose::STANDARD, Engine};
 use libcsrf::{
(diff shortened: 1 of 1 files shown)


horazont:
Thanks for merging and for the fast release!

Issue #948, opened 2025-01-20
openapi: method with same ULR and different path parameters name result in 405


robko23:
## Expected Behavior
Both endpoints should return 200 OK

## Actual Behavior
When having two endpoints on the same URL with different path parameters names result in one of them returning 200 OK and the other one 405 Method Not Allowed. It seems that the order of the functions does not matter, the GET method always gets the 405.
This issue is related to #429 but since that issue is closed and I cannot reopen it I'am opening a new one.

## Steps to Reproduce the Problem
I have created a minimal reproduction of this issue repository: https://github.com/robko23/poem-openapi-routing-issue/tree/master

## Specifications

- Version: poem-openapi@5.1.5 poem@3.1.6
- Platform: GNU/Linux 6.12
- Subsystem: N/A

robko23:
So as I was investigating further, I found out that the issue #429 was fixed by 78fb1123822acf2903fc795879c4a83c78fb448f.
Then there was another issue ( #489 ) which was resolved by 9876172c44a32d0c5e353c4f5eb646a5d2870718.
However these issues seem to be mutually exclusive and the latter commit reverted the fix by the former commit.

zexa:
I was playing around with your code, and it seems that the first method called is the one that will always be allowed through. 

I.e. if you restart the application and call the GET endpoint first, GET will always go through, but if you restart the app and call the post endpoint first, the POST endpoint will always go through.

robko23:
So I have an idea that this could be resolved by renaming the path to normalize param names.
eg. `/hello/:param` -> `/hello/:param0`, `/hello/:another_param` -> `/hello/:param0`.

This would create same entry in the routing table and it could resolve the issue. It could be "injecting" `#[oai(name = "param{i}")]` in the function signature.

I can look into this on the weekend and send PR to resolve this, but a small guidance would be appreciated.

I tried inserting `format!("param{i}")` here, but that did not work.
https://github.com/poem-web/poem/blob/658f7de3106b96941449711eb67fcdc157f1a16e/poem-openapi-derive/src/utils.rs#L127-L132

Then I tried changing this
https://github.com/poem-web/poem/blob/658f7de3106b96941449711eb67fcdc157f1a16e/poem-openapi-derive/src/api.rs#L278-L282

like this
```rust
let param_name = format!("param{i}");
```
but that didn't work either.
However I think that I'm on the right track.
I just need to rename the entry in the routing table and then change this
[...]

Issue #429, opened 2022-11-20
openapi: methods with different argument types at the same url: 405 method not allowed


miquels:
First, let me tell you that I am impressed by poem. I'm working with poem-openapi and it's really good! :), Anyway, on to the issue.

## Expected Behavior
It would be working

## Actual Behavior
405 method not allowed

## Steps to Reproduce the Problem
See https://github.com/miquels/poem , in examples/openapi/users-crud/src/main.rs
Full url: https://github.com/miquels/poem/blob/master/examples/openapi/users-crud/src/main.rs

A bunch of handlers for the same url - /api/users/<something> - with different methods are defined. This works fine in the example code. But if the arguments are of a different type, in the example for `get` `Path<String`, and for `put` `Path<i64>`, then the `put` method will never be called. It results in a `405 method not allowed`.

## Specifications

- Version: 1.3.48
- Platform: freebsd
- Subsystem:

miquels:
I need to add to this that if I define both a `#[oai(path = "/users", method = "get", tag = "ApiTags::User")]` and a `#[oai(path = "/users/:user_name", method = "get", tag = "ApiTags::User")]`, that calling the second one also results in a 405 method not allowed.

miquels:
Still investigating. It looks like the problem is not different arguments, but the _name_ of the argument. If they are all called `user_id` event with different types, it works. If some of them are called `user_id` and one `username` that's what creates the problem.

sunli829:
I'm investigating

sunli829:
Sorry, this is a bug and I'm working on fixing it. 😂

bruwozniak:
@sunli829 I am seeing the same issue persists in current version. was the fix ever released?

braunwolfgang:
@sunli829 @bruwozniak I observed the same issue, i had different naems for the variable and I got the same 405. After making them the same name, it works.

PR #952, opened 2025-01-26
fix(poem-openapi-derive): Allow different path param names on same route

robko23:
This commits add automatic renaming of path parameters in route table and param extraction.
When having two same routes with different method and different param names, it resulted in one of them returning 405 instead of firing the handler.

This fix is somewhat janky, because it checks if the extracted param type is `Path` and then it renames is to `param{i}`. There could be more edge cases down the line, but for now it passess all the tests.

Fixes: https://github.com/poem-web/poem/issues/948
Related: https://github.com/poem-web/poem/issues/429

### DIFF
--- a/poem-openapi-derive/src/api.rs
+++ b/poem-openapi-derive/src/api.rs
@@ -241,6 +241,8 @@ fn generate_operation(
     let mut params_meta = Vec::new();
     let mut security = Vec::new();
 
+    let mut path_param_count = 0;
+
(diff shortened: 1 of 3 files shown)


robko23:
ci should be fixed by https://github.com/poem-web/poem/pull/953

attila-lin:
Hi still working on it?

robko23:
no, this one is done

Issue #950, opened 2025-01-21
The update of the "csrf" crate (v0.4.2) has broken the Poem code.


rxdiscovery:
Hello,

The CSRF handling code (poem > src > web > csrf.rs) is broken because the "csrf" crate was updated (csrf v0.4.2) , changing the return value of the verify_token_pair function. It now returns Result<(), CsrfError> instead of bool.

```rust
      self.protect.verify_token_pair(&token, cookie)
```

```
   |
62 |     pub fn is_valid(&self, token: &str) -> bool {
   |                                            ---- expected `bool` because of return type
...
78 |         self.protect.verify_token_pair(&token, cookie)
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `bool`, found `Result<(), CsrfError>`
   |
   = note: expected type `bool`
              found enum `std::result::Result<(), CsrfError>`
```

## Specifications

- Version: 3.1.6

PR #953, opened 2025-01-26
fix(poem): Handle richer libcsrf error

robko23:
Update crate csrf to version 0.4.2 broke the CsrfVerifier. This commit accounts for the richer error and adds new function that can propagate that information

Fixes: https://github.com/poem-web/poem/issues/950

### DIFF
--- a/poem/src/web/csrf.rs
+++ b/poem/src/web/csrf.rs
@@ -39,6 +39,18 @@ pub struct CsrfVerifier {
     protect: Arc<AesGcmCsrfProtection>,
 }
 
+/// Enum representing CSRF validation error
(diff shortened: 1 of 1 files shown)


attila-lin:
so the csrf feature does not have test？

how about add some test case for it

robko23:
It does, but I didn't bother changing it, because the change is that small. And if the `is_valid` function is covered by test, the `validate` function is also covered.

https://github.com/poem-web/poem/blob/bdd7eac77f2613e8ea0deedbadc1ab13cb50b4ae/poem/src/middleware/csrf.rs#L282

attila-lin:
thank you

PR #958, opened 2025-02-04
feat(openapi): reflect fallback security scheme in spec

patrickariel:
This makes it so that security schemes with a fallback:

```rust
#[derive(SecurityScheme)]
enum MySecurityScheme {
    MySecuritySchemeBasic(MySecuritySchemeBasic),
    #[oai(fallback)]
    NoAuth,
}
```

Will be reflected in the OpenAPI schema:

```jsonc
"security": [
  {
    "MySecuritySchemeBasic": []
  },
  {} // <-- the fallback scheme
]
```

This is the correct way to define optional security schemes, according to the [official specification](https://github.com/OAI/OpenAPI-Specification/blob/d237e681c2bce77413f0dfc4278e168d6ed0edb0/versions/3.0.3.md#optional-oauth2-security) as of version `3.0.3`.

Closes https://github.com/poem-web/poem/issues/959

### DIFF
--- a/poem-openapi-derive/src/api.rs
+++ b/poem-openapi-derive/src/api.rs
@@ -411,6 +411,9 @@ fn generate_operation(
                         (security_name, ::std::vec![#(#crate_name::OAuthScopes::name(&#scopes)),*])
                     ]));
                 }
+                if <#arg_ty as #crate_name::ApiExtractor>::has_security_fallback() {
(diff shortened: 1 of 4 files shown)


stevenj:
This looks great.  Would love to see this merged.

attila-lin:
Thank you

PR #961, opened 2025-02-06
fix(poem-openapi): handle additional_properties correctly in flatten

cristicbz:
Fixes the generated documentation for a type like
```rust
#[derive(Object)]
struct SomeObject {
    fixed: u32,

    #[oai(flatten)]
    items: HashMap<String, u64>
}
```

Currently, such a type shows no documentation for `items`, because `HashMap<String, ...>` only contains `additionalProperties` in the schema. This PR adds tracking of `additionalProperties` in addition to `properties` and `required` when using `flatten`.

In principle we can only support one flattened type with `additionalProperties`, but it wasn't clear to me how to enforce this, so instead this PR just keeps track of the `additionalProperties` of the LAST `flatten`-ed field.

### DIFF
--- a/poem-openapi-derive/src/object.rs
+++ b/poem-openapi-derive/src/object.rs
@@ -108,6 +108,7 @@ pub(crate) fn generate(args: DeriveInput) -> GeneratorResult<TokenStream> {
     let mut register_types = Vec::new();
     let mut fields = Vec::new();
     let mut meta_fields = Vec::new();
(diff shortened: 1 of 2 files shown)


attila-lin:
Hi pls do the fmt. Thank you!

cristicbz:
Done! Sorry, I do run `fmt` on save, but on stable, not `nightly`, so it was not picking up `rustfmt.toml`. Please take another look!

attila-lin:
Thank you.

PR #963, opened 2025-02-07
refactor: middleware transform can take ownership

tisonkun:
This is a breaking change. But since we always use `middleware.transform` exactly once, take the ownership can reduce some unnecessary wrapper.

For example, a middleware can take an owned value, but since `transform` use `&self`, when decorate `endpoint`, you can only pass a ref of that owned value, or copy, or Arc(Mutex).

cc @sunli829 what do you think?


tisonkun:
@sunli829 any follow-up plan?

sunli829:
This is a breaking change, and the need to transfer ownership of internal variables in middleware is rarely seen. For this reason, I am unable to accept this PR. Nevertheless, I sincerely appreciate the effort you put into it.

PR #964, opened 2025-02-08
Support rename in NewType

ihor-rud:
Parametrizing a generic object with two different `NewType`s that have the same inner type leads to a name clash at runtime. This merge request adds a rename parameter to override the default NewType names.

```rust
#[derive(NewType)]
struct NewTypeA(String);

#[derive(NewType)]
struct NewTypeB(String);

#[derive(Debug, Clone, Object)]
pub struct Wrapper<T>
where
    T: Type + ParseFromJSON + ToJSON,
{
    pub id: T,
}

// name clash
assert_eq!(Wrapper::<NewTypeA>::name(), Wrapper::<NewTypeB>::name()); 
```

### DIFF
--- a/poem-openapi-derive/src/newtype.rs
+++ b/poem-openapi-derive/src/newtype.rs
@@ -8,7 +8,7 @@ use quote::quote;
 use syn::{Attribute, DeriveInput, Error, Generics, Type};
 
 use crate::{
-    common_args::ExternalDocument,
+    common_args::{ExternalDocument, LitOrPath},
     error::GeneratorResult,
     utils::{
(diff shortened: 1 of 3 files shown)


attila-lin:
Maybe a path example?

ihor-rud:
> Maybe a path example?

what do mean?

attila-lin:
The code seems support use the `LitOrPath`, now looks like the test only run the `Lit` part

ihor-rud:
> The code seems support use the `LitOrPath`, now looks like the test only run the `Lit` part

I see. I added test.

Issue #965, opened 2025-02-09
Define a range of error codes


musjj:
## Description of the feature

From the [spec](https://swagger.io/docs/specification/v3_0/describing-responses/#http-status-codes):

> To define a range of response codes, you may use the following range definitions: 1XX, 2XX, 3XX, 4XX, and 5XX. If a response range is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code. Each response status requires a description. For example, you can describe the conditions for error responses. Markdown (CommonMark) can be used for rich text representation.
> 
> ```yaml
> responses:
>   "200":
>     description: OK
>   "400":
>     description: Bad request. User ID must be an integer and larger than 0.
>   "401":
>     description: Authorization information is missing or invalid.
>   "404":
>     description: A user with the specified ID was not found.
>   "5XX":
>     description: Unexpected error.
> ```

## Code example (if possible)

```rust
#[derive(ApiResponse)]
pub enum ApiError {
    #[oai(status = 400)]
    BadUserId,
    #[oai(status = 5XX)]
    UnexpectedError(#[status] http::StatusCode),
}
```

PR #966, opened 2025-02-12
Define a range of response codes

patrickariel:
This makes it possible to define a range of response codes, like this:

```rust
use poem::http::StatusCode;
use poem_openapi::{payload::PlainText, ApiResponse};

#[derive(ApiResponse)]
enum CreateUserResponse {
    #[oai(status_range = "2XX")]
    Ok(StatusCode, PlainText<String>),
    #[oai(status_range = "4XX")]
    ClientError(StatusCode),
    #[oai(status_range = "5XX")]
    ServerError(StatusCode),
}
```

The inner `StatusCode` will be used to set the status code of the actual response.

The range syntax is defined in the [official specification](https://github.com/OAI/OpenAPI-Specification/blob/8e3015a1c549375dd64395606cf8f991b0aabbd5/versions/3.1.1.md#patterned-fields-1) as follows:

> To define a range of response codes, this field MAY contain the uppercase wildcard character `X`. For example, `2XX` represents all response codes between `200` and `299`. Only the following range definitions are allowed: `1XX`, `2XX`, `3XX`, `4XX`, and `5XX`.

Closes https://github.com/poem-web/poem/issues/965.

### DIFF
--- a/examples/openapi/content-type-accept/src/bcs_payload.rs
+++ b/examples/openapi/content-type-accept/src/bcs_payload.rs
@@ -84,6 +84,7 @@ impl<T: Serialize + Type> ApiResponse for Bcs<T> {
             responses: vec![MetaResponse {
                 description: "",
                 status: Some(200),
+                status_range: None,
                 content: vec![MetaMediaType {
                     content_type: Self::CONTENT_TYPE,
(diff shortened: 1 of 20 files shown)


attila-lin:
Thank you

PR #971, opened 2025-02-17
Update CI badge in README to show only `master`

chriskrycho:
Previously, this was showing the CI state for *any* branch, which meant that whatever CI run happened to go most recently would be the status reported by the badge. Using [the branch parameter][docs] to use only the `master` branch going forward will provide a more accurate picture to folks looking at the project.

[docs]: https://github.com/github/docs/blob/main/content/actions/monitoring-and-troubleshooting-workflows/monitoring-workflows/adding-a-workflow-status-badge.md#using-the-branch-parameter

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 <div align="center">
   <!-- CI -->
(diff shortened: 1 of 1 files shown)


attila-lin:
thank you

PR #975, opened 2025-02-18
feat: add `Partitioned` option for `CookieConfig`

Cinea4678:
### Summary

This commit adds the `Partitioned` option to the `CookieConfig` object in the built-in Session support of poem.

### Background

Third-party cookies may soon be restricted by Chromium-based browsers (https://goo.gle/3pc-dev-issue). These browsers require developers to use "Cookies Having Independent Partitioned State (CHIPS)" (https://developer.mozilla.org/en-US/docs/Web/Privacy/Privacy_sandbox/Partitioned_cookies), which involves adding the `Partitioned` attribute in `Set-Cookie`. As a result, developers using poem to build applications may need to use the `Partitioned` option in cookies.

### DIFF
--- a/poem/src/session/cookie_config.rs
+++ b/poem/src/session/cookie_config.rs
@@ -26,6 +26,7 @@ pub struct CookieConfig {
     http_only: bool,
     max_age: Option<Duration>,
     same_site: Option<SameSite>,
+    partitioned: bool,
 }
 
 impl Default for CookieConfig {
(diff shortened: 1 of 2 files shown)


attila-lin:
Need a way to get the value?

Cinea4678:
> Need a way to get the value?

Sure, I've added it in the commit

PR #977, opened 2025-02-20
fix: update OpenTelemetry to v0.28

SimenB:
They're targeting a v1 in the summer, which should reduce the churn on this dep 😀 

https://github.com/open-telemetry/opentelemetry-rust/blob/ac69af6a0aea3e2a7a7ad321496ccf3314e5bf64/docs/migration_0.28.md#notes-on-breaking-changes-and-the-path-to-10


attila-lin:
nice! Thank you

PR #980, opened 2025-02-21
feat(static-files): allow setting `no-cache` header on index file

SimenB:
We had an issue where the browser would cache the `index.html` file and then request JS files that was no longer served, resulting in a 404 and the browser erroring out

image

I added this to my own code, but having the capability built in would (besides letting me have less code on my side) make the case discoverable for others and avoid the issue to begin with.

```rust
StaticFilesEndpoint::new(static_assets_path)
    .index_file("index.html")
    .fallback_to_index()
    .around(|ep, req| async move {
        let req_path = req.uri().path();

        if req_path != "/" && req_path != "/index.html" {
            return ep.call(req).await;
        }

        let mut resp = ep.call(req).await?;

        if !resp.is_success() {
            return Ok(resp);
        }

        resp.headers_mut()
            .entry(CACHE_CONTROL)
            .or_insert(HeaderValue::from_static("no-cache"));

        Ok(resp)
    })
```

(looking at it I should probably check if the server requested `html` rather than the path... but this works for us as the way the app is hosted, index is _always_ requested first)

### DIFF
--- a/poem/src/web/static_file.rs
+++ b/poem/src/web/static_file.rs
@@ -38,6 +38,8 @@ pub enum StaticFileResponse {
         last_modified: Option<String>,
         /// `Content-Range` header value
         content_range: Option<(std::ops::Range<u64>, u64)>,
+        /// `Cache-Control` header value
+        cache_control: Option<String>,
     },
     /// 304 NOT MODIFIED
     NotModified,
@@ -63,6 +65,7 @@ impl IntoResponse for StaticFileResponse {
                 etag,
                 last_modified,
                 content_range,
+                cache_control,
             } => {
                 let mut builder = Response::builder()
                     .header(header::ACCEPT_RANGES, "bytes")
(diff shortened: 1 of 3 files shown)


poem/src/web/static_file.rs
```diff
@@ -167,17 +169,23 @@ impl StaticFileRequest {
...
     ///
-    /// `prefer_utf8` - Specifies whether text responses should signal a UTF-8
-    /// encoding.
+    /// # Arguments
+    ///
+    /// * `prefer_utf8` - Specifies whether text responses should signal a UTF-8
+    ///   encoding.
+    /// * `no_cache` - Specifies whether to set the `Cache-Control` header to
+    ///   `no-cache`.
     pub fn create_response(
```
SimenB, line 184: might be that it would be cleaner with and options bag rather than two bool arguments. didn't think about the fact this was a breaking change to begin with 😅

PR #986, opened 2025-02-22
fix(openapi): do not use `cookie` feature by default

SimenB:
I do not use this feature, and it's the only thing that pulls in `chrono` in my project. Is it really needed for `poem-openapi` to work?


attila-lin:
will land in next big version

Thank you

PR #989, opened 2025-02-23
fix(openapi): exclude style parameter when serializing if none

snormore:
Update `MetaOperationParam.style` to `skip_serializing_if = "Option::is_none"` so that `"style": null` isn't included in the serialized spec output, which isn't a valid style value.


attila-lin:
close #992

PR #997, opened 2025-03-06
fix: guard cookie features behind feature toggle

SimenB:
I'm not sure why CI is happy, but running `cargo test` fails to compile on master, and presumably I broke it in https://github.com/poem-web/poem/pull/986.

With the feature gates added here, both `cargo test` and `cargo test -F cookie` passes on my machine


attila-lin:
Thankyou！

PR #999, opened 2025-03-08
`from_request()` error handling

stella3d:
## `explode = false` panic
we have an issue in our servers that has prevented us from using `explode = false` to accept array parameters in routes.

when a request was made missing some expected query parameters with `explode = false`, the request would panic on the second `.unwrap()` call in the previous version of `from_request`.

this adds some basic error handling around that scenario to prevent runtime panics.

### `"cookie"` feature change
i'm not sure why, this was just necessary to get it building. unrelated to the real change. seems to be an issue on master as well.

### DIFF
--- a/poem-openapi/Cargo.toml
+++ b/poem-openapi/Cargo.toml
@@ -35,6 +35,7 @@ poem = { workspace = true, default-features = true, features = [
   "sse",
   "xml",
   "yaml",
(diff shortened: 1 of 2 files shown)


attila-lin:
Hi cookie should fix now. would you remove it again and fmt？

PR #1002, opened 2025-03-14
chore: add track_caller to test asserts

luca-iachini:
This PR adds the `track_caller` attribute to synchronous test assertions, so errors point to the test call site instead of the library's code. Since `track_caller` isn't supported for async methods, this at least improves error reporting for sync assertions, making debugging easier.


attila-lin:
LGTM
Thank you.

PR #1016, opened 2025-04-09
openapi: support `Duration` and `Timestamp` from `prost_wkt_types`

Shyrogan:
This pull request adds support for new types in `prost_wkt_types`.

PR #1017, opened 2025-04-09
feat(grpc-build): allow calling prost's `external_path` function

Shyrogan:
This is a follow up to #1016 which allows for serialization of `prost-wkt-types`.

In order to use the `Timestamp` type, you have to use add the following code to the `build.rs` (see https://github.com/tokio-rs/prost/issues/672):
```rust
        .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
```
which is not possible in current `poem-grpc-build` config.

### DIFF
--- a/poem-grpc-build/src/config.rs
+++ b/poem-grpc-build/src/config.rs
@@ -300,6 +300,16 @@ impl Config {
         self
     }
 
+    /// Declare an externally provided Protobuf package or type.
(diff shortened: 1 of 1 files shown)


gwik:
Hi @Shyrogan, could you also add `compile_well_known_types`? so I can close my PR. https://github.com/poem-web/poem/pull/1025/files#diff-b6e0d2aa6727a4b12527be5922af5ab0ad36aa5ee1ddbac375d24164a60a80a0R303-R309

PR #1019, opened 2025-04-10
feat(openapi): add scalar ui support

Pribess:
Add scalar ui support

## Example code
```
use poem::{Route, listener::TcpListener};
use poem_openapi::{OpenApi, OpenApiService, param::Query, payload::PlainText};

struct Api;

#[OpenApi]
impl Api {
    #[oai(path = "/hello", method = "get")]
    async fn index(&self, name: Query<Option<String>>) -> PlainText<String> {
        match name.0 {
            Some(name) => PlainText(format!("hello, {}!", name)),
            None => PlainText("hello!".to_string()),
        }
    }
}

#[tokio::main]
    let api_service =
        OpenApiService::new(Api, "Hello World", "1.0").server("http://localhost:3001/api");
    let ui = api_service.scalar();
    let app = Route::new().nest("/api", api_service).nest("/", ui);

    poem::Server::new(TcpListener::bind("0.0.0.0:3001"))
        .run(app)
        .await
}
```
Screenshot 2025-04-10 at 23 43 59


attila-lin:
Thank you

PR #985, opened 2025-02-22
chore: bump `rustls`, use `aws_lc_rs` instead of `ring`

attila-lin:

PR #1021, opened 2025-04-18
refactor: populate default provider

tisonkun:
cc @sunli829 

This refers to #985 and imporve usability.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


tisonkun:
```
failures:

---- listener::native_tls::tests::tls_listener stdout ----

thread 'listener::native_tls::tests::tls_listener' panicked at poem/src/listener/native_tls.rs:201:59:
called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:../crypto/evp/evp_fetch.c:349:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Weird. Should not accept this case and I can't reproduce locally.

sunli829:
thanks!

tisonkun:
@sunli829 Do you have some ideas why the referred case failed above?

PR #1031, opened 2025-05-08
Fix webhook nesting

terhechte:
The generation of webhook metadata is not according to the specs. Here is an example of a spec compliant webhook method:

``` yaml
webhooks:
  newPet:
    post:
      requestBody:
     ...
```

Whereas the current poem implementation generates this:

```yaml
webhooks:
  newPet:
    requestBody:
```

E.g. it omits the `method` parameter. This PR fixes that.
In addition, I also improved the documentation. Both Rapidoc and Swagger UI require a `description` to display a webhook. Descriptions are not provided as tags but as Rust docs. This behaviour is not apparent from the Webhook example code.


attila-lin:
Thank you!

PR #1037, opened 2025-05-21
feat: allow calls Path's and Query's from_request with necessary arg in sync context

tisonkun:
cc @sunli829


tisonkun:
We have `Request::path_params` and `Request::params` already.

PR #1041, opened 2025-05-27
add support for num::NonZero

ihor-rud:
This mr implements Type, To* and From* traits for NonZero type wrapper. I as well tried my best to express non zero requirement in openapi model.

Btw I noticed that name for integer types are generated as `integer_{}` but for unsigned as `integer({})`. I am not sure if its intentional but I kept that for NonZero.


attila-lin:
Thank you

PR #1043, opened 2025-05-31
Add support to externally tagged unions

luca-iachini:
This PR adds support for externally tagged unions, which is the default representation used by serde.
I'm currently adding OpenAPI specifications to APIs that accept JSON in this format. Since I can't change the existing JSON payloads, this change makes it possible to generate the OpenAPI spec simply by setting the `externally_tagged` attribute.
(I chose the `externally_tagged` attribute name, but I'm open to renaming it if a better alternative comes up. 🙂)


sunli829:
Thank you, this seems to provide some convenience in certain situations.

PR #1049, opened 2025-06-05
Support deep linking in Swagger UI

MagnusHodne:
Adds support for [deep linking](https://swagger.io/docs/open-source-tools/swagger-ui/usage/deep-linking/) in Swagger UI, which makes it easier to keep API tabs/dropdowns open when refreshing the web page.

Would be very nice to have something like a `custom_swagger_ui(swagger_conf)` function added, so that deepLinking, filter and other properties can be configured manually. I'm happy to try my hand at implementing something like that if desirable :)


attila-lin:
> Would be very nice to have something like a `custom_swagger_ui(swagger_conf)` function added, so that deepLinking, filter and other properties can be configured manually. I'm happy to try my hand at implementing something like that if desirable :)

welcome

PR #1075, opened 2025-07-20
Feat: Add cloudflare worker support using poem

tiannian:
By making minor modifications to Poem (modification of 1 function and 1 dependency), Poem can be compiled to wasm and some Cloudflare related APIs are provided. Now we can use poem to write cloudflare worker.

Note: I separated `TokioIo` from `hyper-util` to avoid using this type to import `tokio/net`. I also submitted a PR to `hyper-util`, but I don't know when it can be merged.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,8 @@ members = [
     "poem-grpc-build",
     "poem-grpc",
     "poem-mcpserver",
-    "poem-mcpserver-macros", "poem-worker",
+    "poem-mcpserver-macros",
+    "poem-worker",
 ]
 
 [workspace.package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ members = [
     "poem-grpc-build",
     "poem-grpc",
     "poem-mcpserver",
-    "poem-mcpserver-macros",
+    "poem-mcpserver-macros", "poem-worker",
 ]
 
 [workspace.package]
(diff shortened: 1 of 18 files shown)


sunli829:
thanks, looks good to me.  🙂

sunli829:
It doesn't compile, could you help me? @tiannian 

```
error[E0308]: mismatched types
  --> poem-worker\src\cloudflare.rs:14:9
   |
13 |     pub fn asn(&self) -> u32 {
   |                          --- expected `u32` because of return type
14 |         self.0.asn()
   |         ^^^^^^^^^^^^ expected `u32`, found `Option<u32>`
   |
   = note: expected type `u32`
              found enum `std::option::Option<u32>`
help: consider using `Option::expect` to unwrap the `std::option::Option<u32>` value, panicking if the value is an `Option::None`
   |
14 |         self.0.asn().expect("REASON")
   |                     +++++++++++++++++

error[E0308]: mismatched types
  --> poem-worker\src\cloudflare.rs:18:9
   |
17 |     pub fn as_organization(&self) -> String {
   |                                      ------ expected `std::string::String` because of return type
18 |         self.0.as_organization()
   |         ^^^^^^^^^^^^^^^^^^^^^^^^ expected `String`, found `Option<String>`
   |
   = note: expected struct `std::string::String`
                found enum `std::option::Option<std::string::String>`
help: consider using `Option::expect` to unwrap the `std::option::Option<std::string::String>` value, panicking if the value is an `Option::None`
   |
18 |         self.0.as_organization().expect("REASON")
   |                                 +++++++++++++++++
```

PR #1077, opened 2025-07-28
Fix cloudfare `worker-rs` new version changes

tiannian:
Caused by https://github.com/cloudflare/workers-rs/releases/tag/v0.6.1 changes, we need add `Option` type on `Cloudflare`'s object.


tiannian:
I don't understand why there is a clippy error on CI. My local seems to be normal. @sunli829

sunli829:
thanks

PR #1082, opened 2025-08-05
fix(deps): add macros feature to time dependency

cp90-pixel:
Resolves compilation error when time feature is enabled:
- Add "macros" feature to time dependency in Cargo.toml
- Fixes unresolved import `time::macros` error in poem-openapi

Fixes build failure when using poem-openapi with time feature enabled.


attila-lin:
thank you

attila-lin:
thank you

PR #1083, opened 2025-08-05
feat(openapi): add `Type`, `To*`, and `ParseFrom*` implementations for `unit` type

mattremmel:
Motivation
---
Implementing patterns such as uniform response, as demonstrated in `examples/openapi/uniform-response`, where a response type generic over `T` is used to return `T` in a data field is awkward when unit type is desired. For cases where there is no meaningful `T`, without being able to use `()` as `T`, one is forced to use work arounds such as `#[derive(Object)] struct Unit;` or `types::Any<()>`. These changes will allow using `()` directly for those use cases. The deserialization and serialization behavior matches the behavior of serde_json's behavior for unit types.

Modifications
---
- Add `Type`, `ParseFromJSON`, `ParseFromParameter`, `ParseFromMultipartField`, `ToJSON`, and `ToHeader` trait implementation for unit type.
- Update OpenAPI uniform response example to use unit type in MyResponse to demonstrate use case.

### DIFF
--- a/examples/openapi/uniform-response/src/main.rs
+++ b/examples/openapi/uniform-response/src/main.rs
@@ -85,6 +85,12 @@ impl Api {
         *self.resource.lock().await = Some(obj.0);
         MyResponse::Ok(Json(ResponseObject::ok(true)))
     }
+
+    #[oai(path = "/resource", method = "post")]
+    async fn post(&self, obj: Json<Resource>) -> MyResponse<()> {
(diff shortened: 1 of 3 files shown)


attila-lin:
Nice!

Would you mind some test case?

mattremmel:
Yea not a problem. I'll add those as soon as I can this week.

attila-lin:
Thank you!

PR #1093, opened 2025-09-06
feat(poem): allow customization of any HTTP2 parameters

DDtKey:
### Description

This PR introduces ability to customize any HTTP2 server params via passing "customizer" closure which can mutate the H2 builder.

This implementation also will not require to expose all parameters manually, but instead provides a flexible way to configure h2 params as they wish.

For instance, there is a lack of ability to configure http2_adaptive_window and related configs.

### Example

```rs
            .http2_builder_customize(|builder| {
                builder.adaptive_window(true)
                    .max_concurrent_streams(1000);
            })

```

### DIFF
--- a/poem/src/server.rs
+++ b/poem/src/server.rs
@@ -15,7 +15,7 @@ use std::{
 use futures_util::FutureExt;
 use http::uri::Scheme;
 use hyper::body::Incoming;
-use hyper_util::server::conn::auto;
+use hyper_util::server::conn::auto::{self, Http2Builder};
 use pin_project_lite::pin_project;
 use tokio::{
     io::{AsyncRead, AsyncWrite, ReadBuf, Result as IoResult},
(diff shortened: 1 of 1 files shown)


poem/src/server.rs
```diff
@@ -57,6 +58,7 @@ impl<L: Listener> Server<L, Infallible> {
             http2_max_concurrent_streams: None,
             http2_max_pending_accept_reset_streams: Some(20),
             http2_max_header_list_size: 16384,
```
DDtKey, line 60: we could deprecate these parameters in general, but it's up to you - perhaps in the future

poem/src/server.rs
```diff
@@ -270,6 +290,19 @@ where
     }
 }
 
+trait Http2ConfigCustomize: Send + Sync {
+    fn customize(&self, builder: &mut Http2Builder<'_, hyper_util::rt::TokioExecutor>);
+}
```
DDtKey, line 295: It’s for simplicity, and it’s also possible to allow passing not only closures, but custom implementations of the customizer if needed

PR #1100, opened 2025-09-27
feat(openapi): add `Type`, `To*`, and `ParseFrom*` implementations for `std::path::PathBuf` and `std::path::Path`

0xb002f0:
Adds Type, ParseFromJSON, ParseFromParameter, ParseFromMultipartField, ToJSON, and ToHeader trait implementations for `std::path::PathBuf` and `std::path::Path`.


attila-lin:
Thank u！

PR #1120, opened 2025-12-05
feat(openapi): Add Geometry type support and fix array schema validation

gauteoh:
Fix OpenAPI schema generation for arrays and enhance GeoJSON support

This PR adds better support for GeoJSON types in poem-openapi.

**Changes**

**Fixed array schema validation fields**
- Changed fixed-length arrays to use min_items and max_items instead of min_length and
max_length
- The OpenAPI spec defines minItems/maxItems for arrays, while minLength/maxLength are for
strings
- Affects: poem-openapi/src/types/external/array.rs:24-25

**Enhanced GeoJSON type schemas**
- Added enum constraint to the type field so it validates against the specific geometry type
(e.g., "Point", "LineString") rather than accepting any string
- Added bbox field to schemas with proper validation (optional array of numbers with minimum
4 items)
- Added descriptive titles to all GeoJSON schemas for better API documentation

**Implemented full Geometry enum support**
- Added Type, ParseFromJSON, and ToJSON trait implementations for geo_types::Geometry
- Schema uses oneOf to represent the union of all geometry types (Point, LineString, Polygon,
 MultiPoint, MultiLineString, MultiPolygon)
- Enables using Geometry directly in API endpoints and data structures

**Test coverage**

Added tests covering:
- Schema structure validation for type enums, bbox fields, and proper use of array validation
 fields
- Serialization and deserialization for Geometry enum variants
- Schema oneOf structure for Geometry union type

### DIFF
--- a/poem-openapi/src/types/maybe_undefined.rs
+++ b/poem-openapi/src/types/maybe_undefined.rs
@@ -83,20 +83,15 @@ use crate::{
 /// cli.get("/get").send().await.assert_json(json!({"attr1": null, "attr2": "abc"}));
 /// # });
 /// ```
-#[derive(Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Debug, Hash)]
+#[derive(Copy, Clone, PartialEq, PartialOrd, Eq, Ord, Debug, Hash, Default)]
 pub enum MaybeUndefined<T> {
     /// Undefined
+    #[default]
     Undefined,
     /// Null
     Null,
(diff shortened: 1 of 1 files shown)


gauteoh:
Now the checks are successful, there is however an error in poem-openapi/src/types/maybe_undefined.rs:96:1 from clippy, which I guess should be resolved in another PR?

attila-lin:
Thank you！

Issue #945, opened 2025-01-05
openapi Union does not properly use flatten in documentation


rsfunc:
## Expected Behavior

For redoc for flatten endpoint I expect some help about the fields:

```{
"other": "flatten",
"type": "ValueB",
"value": true
}
```

NOTE: the actual output of the endpoint is correct, only the documentation is not!

## Actual Behavior

![image](https://github.com/user-attachments/assets/714cb0b4-337a-4071-987d-d6965c847037)

## Steps to Reproduce the Problem

1. Compile the given sample

Cargo.toml
```
[package]
name = "example-openapi-flatten"
version.workspace = true
edition.workspace = true
publish.workspace = true

[dependencies]
poem = { version="3.1.6" }
poem-openapi = { version="5.1.5", features= ["redoc", "swagger-ui"]}
tokio = { version = "1.42.0", features = ["macros", "rt-multi-thread"] }
tracing-subscriber = {version = "0.3.19" }
```

Main.rs

```
use poem::{listener::TcpListener, Route, Server};
use poem_openapi::{
    param::Query,
    payload::{Json, PlainText},
    Object, OpenApi, OpenApiService, Union,
};

struct Api;

#[derive(Object)]
struct ValueA {
    pub value: bool,
}

#[derive(Object)]
struct ValueB {
    pub value: bool,
}

#[derive(Union)]
#[oai(one_of = true, discriminator_name = "type")]
enum UnionEnum {
    ValueA(ValueA),
    ValueB(ValueB),
}

#[derive(Object)]
struct Flatten {
    #[oai(flatten)]
    inner: UnionEnum,
    other: String,
}

#[OpenApi]
impl Api {
    #[oai(path = "/flatten", method = "get")]
    async fn flatten(&self) -> Json<Flatten> {
        Json(Flatten {
            inner: UnionEnum::ValueB(ValueB { value: true }),
            other: "flatten".into(),
        })
    }
}

#[tokio::main]
async fn main() -> Result<(), std::io::Error> {
    if std::env::var_os("RUST_LOG").is_none() {
        std::env::set_var("RUST_LOG", "poem=debug");
    }
    tracing_subscriber::fmt::init();

    let api_service =
        OpenApiService::new(Api, "Hello World", "1.0").server("http://localhost:3000/api");
    let ui = api_service.redoc();

    Server::new(TcpListener::bind("0.0.0.0:3000"))
        .run(Route::new().nest("/api", api_service).nest("/", ui))
        .await
}
```

[...]

PR #1144, opened 2025-12-21
fix: properly generate OpenAPI schema for flattened Union types

coleleavitt:
## Summary
- Fixes issue #945: OpenAPI Union does not properly use flatten in documentation

## Problem
When a Union type (using `oneOf`/`anyOf`) was flattened into an Object, the generated OpenAPI schema was incorrect. The Object derive macro was trying to extend `properties` from the Union's schema, but Union schemas have `oneOf`/`anyOf` instead of `properties`.

## Solution
The fix detects when a flattened field is a Union type and uses `allOf` composition instead:

```yaml
# Before (broken):
Flatten:
  type: object
  properties:
    other:
      type: string
  # Missing the Union fields!

# After (correct):
Flatten:
  allOf:
    - $ref: '#/components/schemas/UnionEnum'  # The Union
    - type: object
      properties:
        other:
          type: string
```

## Changes
- Added `is_union()` helper method to `MetaSchema` to detect Union types
- Modified Object derive macro to check if flattened types are Unions
- For Union types: adds the Union reference to `allOf`
- For regular Object types: continues to extend properties as before

## Example Usage
```rust
#[derive(Union)]
#[oai(discriminator_name = "type")]
enum UnionEnum {
    ValueA(ValueA),
    ValueB(ValueB),
}

#[derive(Object)]
struct Flatten {
    #[oai(flatten)]
    inner: UnionEnum,  // Now properly generates allOf schema
    other: String,
}
```

### DIFF
--- a/poem-openapi-derive/src/object.rs
+++ b/poem-openapi-derive/src/object.rs
@@ -317,13 +317,31 @@ pub(crate) fn generate(args: DeriveInput) -> GeneratorResult<TokenStream> {
             });
         } else {
             meta_fields.push(quote! {
-                fields.extend(registry.create_fake_schema::<#field_ty>().properties);
+                let flattened_schema = registry.create_fake_schema::<#field_ty>();
+                // Check if the flattened type is a Union (uses oneOf/anyOf)
+                if flattened_schema.is_union() {
+                    // For Union types, add to all_of for proper schema composition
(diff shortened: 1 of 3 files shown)


rsfunc:
Exciting news ! Hopefully it will be merged into master.

PR #1157, opened 2026-01-22
fix(poem): compression middleware

DDtKey:
Improve Accept-Encoding negotiation in compression middleware: fix zstd typo, add explicit `identity` handling, and make `*` resolve to the best enabled algorithm instead of hard‑coding zstd, aligning with MDN guidance on [`Accept-Encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding). 

Main changes:
- Fixed typo in `zstd` encoding name (there was `zsdt`)
- Added parsing support for the `identity` encoding (which must be acceptable and leads to no-compression behavior by the spec)
- Updated star (`*`) handling to only accept enabled algorithms
- Refactored the encoding mapping for better structure and readability

### DIFF
--- a/poem/src/middleware/compression.rs
+++ b/poem/src/middleware/compression.rs
@@ -12,6 +12,7 @@ enum ContentCoding {
     Brotli,
     Deflate,
     Gzip,
+    Identity,
     Star,
     Zstd,
 }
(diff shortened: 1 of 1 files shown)


poem/src/middleware/compression.rs
```diff
@@ -26,16 +27,61 @@ impl FromStr for ContentCoding {
...
+    #[inline]
+    fn priority(&self) -> u8 {
+        match *self {
+            ContentCoding::Deflate => 1,
+            ContentCoding::Gzip => 2,
+            ContentCoding::Brotli => 3,
+            ContentCoding::Zstd => 4,
+            _ => 0,
+        }
+    }
```
DDtKey: This method hasn't been changed in fact, just moved to impl block for slightly better structuring and calling (`coding.priority()` instead of `coding_priority(coding)`)

attila-lin:
Thank you！

PR #1159, opened 2026-01-26
feat(poem-openapi) Allow childless enum variant

Nahuel-M:
Allows the use of a childess enum variant, like this:
```rust
    #[derive(Union, Debug, PartialEq)]
    #[oai(discriminator_name = "type")]
    enum MyObj {
        #[oai(mapping = "c")]
        A(A),
        #[oai(mapping = "d")]
        B(B),
        #[oai(mapping = "e")]
        C,
    }
```

Here, variant C is the equivalent of an empty object, or an object only containing the discriminator if there is one. This allows seamless integration between a domain model and openapi spec.

Please note that I opted for creating a  named schema when using a discriminator, and an inline schema elsewise. 
Let me know what you think!


attila-lin:
Thank you!

PR #1174, opened 2026-03-26
fix(cookies): allow cookies and sessions to update on error

paring-chan:
`CookieJar` and `Session` is not able to update cookies if the endpoint returned an error. It makes extractors returning Err cannot guarantee the session or cookie is updated (especially something like flash message).

This change will make cookies can be updated even an error is returned from endpoint


paring-chan:
Closing as I don't have idea to fix this issue without converting error into response, which causes the test failure.

PR #1175, opened 2026-03-27
feat: add examples for optional/anonymous endpoint with security scheme

ericbsantana:
## Summary

This PR adds a tiny example for a possibly common use case: endpoints that should work for both authenticated and anonymous users. Strictly using `SessionAuthorization` (from the example) crashes with `authorization error` before falling into 

On the example, `/hello` is accessible to everyone, but personalizes the request when a valid `session` cookie is present. For `/protected` route, still strict authentication. Docs were also updated on `poem-openapi` `docs/security_scheme.md` file explaining the `#[oai(fallback)]` pattern for optional auth

This example PR might help others who face these issues with anonymous x authenticated routes. If maintainers see any alternative approaches that matches with a change on the `security_schema` cookie parsing to allow "anonymous" users when present, I’d be happy to collaborate on a follow-up PR too!

### DIFF
--- a/examples/openapi/auth-optional-apikey/src/main.rs
+++ b/examples/openapi/auth-optional-apikey/src/main.rs
@@ -49,7 +49,8 @@ struct Api;
 
 #[OpenApi]
 impl Api {
-    /// Returns a personalized greeting when the `session` cookie is present and valid.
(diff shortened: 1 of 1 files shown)


attila-lin:
Thank you!

PR #1177, opened 2026-04-10
perf: set tcp_nodelay - disable naggle algorithm on tcp

glebpom:
## Summary

Sets `TCP_NODELAY` to true by default on accepted connections. This prevents a common networking pitfall that causes significant latency in modern web services.

## The Problem

Without `TCP_NODELAY,` the interaction between Nagle’s Algorithm and TCP Delayed ACKs creates a "stalemate" where the stack waits for a full buffer or an acknowledgement before sending data.

In HTTP/2 and small-payload scenarios, this manifests as a consistent 40ms latency floor for P99 metrics, as the kernel waits for a timer to expire before flushing frames.

## The Fix

Enabling `TCP_NODELAY` ensures that packets are sent immediately, which is the industry standard across modern high-performance web servers, proxies, and runtimes to ensure predictable low latency.

### Performance: 

Eliminated the 40ms "stutter," reducing trivial handler latency from >40ms to <1ms.

### Standardization:

Brings Poem in line with expected defaults for modern networking stacks, where low-latency delivery is prioritized over legacy packet coalescing.

### DIFF
--- a/poem/src/listener/tcp.rs
+++ b/poem/src/listener/tcp.rs
@@ -73,6 +73,8 @@ impl Acceptor for TcpAcceptor {
     #[inline]
     async fn accept(&mut self) -> Result<(Self::Io, LocalAddr, RemoteAddr, Scheme)> {
         self.listener.accept().await.map(|(io, addr)| {
+            let _ = io.set_nodelay(true);
+
             (
                 io,
                 self.local_addr.clone(),


glebpom:
Hey @attila-lin, could you please take a look?

attila-lin:
Done, thank you

PR #1184, opened 2026-05-21
fix: #1182: remove rustls-pemfile dependency

### LINKED ISSUE #1182: RUSTSEC-2025-0134 vulnerability due to unmaintained `rustls-pemfile` crate
[RUSTSEC-2025-0134](https://rustsec.org/advisories/RUSTSEC-2025-0134.html) states that users should no longer depend on `rustls-pemfile`.

BatmanAoD:
Resolves #1182.

Changes were authored by Claude Opus 4.6, but look correct to me.


attila-lin:
Thank u！

PR #1191, opened 2026-08-24
fix(grpc): correct content-type for JsonCodec

### LINKED ISSUE #1190: grpc: JsonCodec emits a non-gRPC response content-type (`application/json`), breaking conforming gRPC clients
## Expected Behavior

When a service is generated with `.codec("poem_grpc::codec::JsonCodec")`, responses should carry a
content-type that conforms to the gRPC HTTP/2 protocol, i.e. `application/grpc[+proto|+json|+custom]`.
The [spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) defines a single
`Content-Type` production (line 40) that `Response-Headers` also references (line 107), so the
response header is bound by the same grammar as the request header.

This is also what `JsonCodec`'s own doc comment promises:

```rust
// src/codec/json_codec/mod.rs:14
/// A [`Codec`] for JSON `application/grpc+json`
pub struct JsonCodec<T, U>(PhantomData<(T, U)>);
```

## Actual Behavior

Responses carry `content-type: application/json`, which is not a valid gRPC content-type.
Conforming clients reject the response even though the body is a perfectly well-formed
length-prefixed gRPC frame:

```
unexpected HTTP status code received from server: 200 (OK);
transport: received unexpected content-type "application/json"; data: "\x00\x00\x0..."
```

The server logs nothing - from its point of view the call succeeded with HTTP 200.

### Cause

The response content-type is taken from `Codec::CONTENT_TYPES[0]`:

```rust
// src/server.rs:42 (also 90, 134, 184)
let mut resp = Response::default().set_content_type(T::CONTENT_TYPES[0]);
```

and the two JSON codecs list the same two strings in **opposite** order:

```rust
// src/codec/json_codec/mod.rs:30  -> JsonCodec
const CONTENT_TYPES: &'static [&'static str] = &["application/json", "application/grpc+json"];
// src/codec/json_codec/mod.rs:90  -> JsonI64ToStringCodec
const CONTENT_TYPES: &'static [&'static str] = &["application/grpc+json", "application/json"];
```

So `JsonI64ToStringCodec` happens to produce a valid response header while `JsonCodec` does not.
The only difference is the ordering, which is very easy to miss - we hit this by switching from
`JsonI64ToStringCodec` to `JsonCodec` for an unrelated reason (we had adopted `pbjson`, which already
serializes 64-bit integers as strings, so the i64 wrapper had become redundant).

For reference, the client-side validation in grpc-go: `grpcutil.ContentSubtype`
([internal/grpcutil/method.go](https://github.com/grpc/grpc-go/blob/master/internal/grpcutil/method.go))
accepts only `application/grpc`, or `application/grpc` followed by `+` or `;`; otherwise
[internal/transport/http2_client.go](https://github.com/grpc/grpc-go/blob/master/internal/transport/http2_client.go)
emits the error above.

1. Take `examples/grpc/helloworld` and add `.codec("poem_grpc::codec::JsonCodec")` to its `build.rs`.
2. Call it from any conforming gRPC client that sends `content-type: application/grpc+json`
   (e.g. grpc-go with a `json` codec registered).
3. The client rejects the response with `received unexpected content-type "application/json"`.
   Swapping the codec to `JsonI64ToStringCodec` makes it work, with no other change.

## Suggested Fix

Two options, not mutually exclusive:

1. **Minimal** - reorder `JsonCodec::CONTENT_TYPES` to `["application/grpc+json", "application/json"]`,
   matching `JsonI64ToStringCodec`. The response header becomes valid while `application/json` is still
   accepted inbound (`check_content_type` scans the whole list).

2. **More correct** - echo the request's content-type in the response instead of hardcoding
   `CONTENT_TYPES[0]`. The spec's worked example (lines 160/172) shows request and response carrying the
   same value, and grpc-go builds its response header from the request's content-subtype
   (`grpcutil.ContentType`). This would additionally let a plain `application/json` caller get
   `application/json` back. Option 1 alone is enough to unbreak gRPC clients.

Either way, a doc comment on `Codec::CONTENT_TYPES` noting that **the first entry becomes the response
header** would help - right now that is only discoverable by reading `server.rs`.

## Specifications

- poem-grpc 0.5.9 (latest on crates.io)
- Client: grpc-go, via an API gateway
- gRPC spec: https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

kingcanfish:
## Problem

`JsonCodec::CONTENT_TYPES` lists `"application/json"` before `"application/grpc+json"`,
unlike its sibling `JsonI64ToStringCodec` which has the correct order. Since the
generated server handlers use `CONTENT_TYPES[0]` as the literal `content-type` header
value for outgoing responses, a service configured with `JsonCodec` sends back
`content-type: application/json` — not a valid gRPC content-type per the
[HTTP/2 protocol spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md).
Conforming gRPC clients (e.g. grpc-go) reject the response outright, even though the
body itself is a well-formed, correctly-encoded gRPC frame.

Reproduced against a real grpc-go client: a request sent with
`content-type: application/grpc+json` is decoded and handled correctly by the server,
but the client rejects the response with:

```
rpc error: code = Unknown desc = unexpected HTTP status code received from server:
200 (OK); transport: received unexpected content-type "application/json"
```

## Fix

1. Swap `JsonCodec::CONTENT_TYPES` to `["application/grpc+json", "application/json"]`,
   matching `JsonI64ToStringCodec`.
2. In `GrpcServer`'s four RPC handlers (unary / client-streaming / server-streaming /
   bidi-streaming), stop relying on `CONTENT_TYPES[0]` as the response content-type.
   Instead, echo the incoming request's `content-type` header back on the response
   (falling back to `CONTENT_TYPES[0]` only when the request has none). This removes
   the whole class of bug — a future custom `Codec` with a wrong `CONTENT_TYPES`
   ordering can no longer break interop the same way.

## Testing

- `cargo test -p poem-grpc --all-features` — all existing tests pass.
- Manually reproduced against a real grpc-go client / poem-grpc server pair: confirmed
  the failure before the fix, and confirmed the RPC succeeds after.

Closes #1190


kingcanfish:
@sunli829 @attila-lin  kindly ping, please review this when you have time

kingcanfish:
update: force-pushed to squash a cargo fmt fixup into the original commit

attila-lin:
Thanks for the fix! CI on this branch is blocked by master rustfmt drift (unrelated comment wraps), and we don't have push access to `kingcanfish/poem` to update the branch.

I've opened a minimal follow-up from the maintainer fork that:
1. keeps the essential `CONTENT_TYPES` reorder
2. includes the rustfmt fixes so CI can pass

Will merge that and close this once green. Credit remains yours for finding and fixing the bug.

attila-lin:
Thanks for the fix — the `JsonCodec::CONTENT_TYPES` reorder looks right and matches `JsonI64ToStringCodec`.

Could you update the branch so CI goes green? Right now checks are failing mainly on:

1. **`cargo fmt --check`** — master currently has a few rustfmt comment-wrap drifts (unrelated files + examples). Please rebase on latest `master` and run `cargo fmt --all` (nightly rustfmt, same as CI).
2. Optionally reconsider echoing the request `Content-Type` back on the response — the `CONTENT_TYPES` reorder alone should fix the grpc-go rejection; echoing the request header is a bit broader than needed (e.g. charset / unexpected subtypes). Happy either way if you prefer keeping it for interop.

Once CI is green we can merge. Thanks again!

kingcanfish:
@attila-lin thanks for the review 
1. Rebased on latest `master`, squashed, ran nightly `cargo fmt --all` (incl. `examples/`).
2. Dropped the `Content-Type` echo — now only reorders `JsonCodec::CONTENT_TYPES`.
3. With fmt fixed, CI then fails at clippy (`-D warnings`): stable 1.98 flags `result_large_err` on every `Result<_, Status>` since `Status` is exactly 128 bytes (`HeaderMap` is 96). Boxed `Status::metadata` (private field, public API unchanged) to bring it down to 40 bytes, and fixed two `needless_borrows_for_generic_args` in `format!` args.

Verified locally with rustc 1.98.1 + `-D warnings`: fmt, clippy (all matrix packages + examples) and tests pass. Same clippy failure your follow-up PR hit.

PR #1197, opened 2026-09-22
fix(grpc): put application/grpc+json first for JsonCodec

### LINKED ISSUE #1190: grpc: JsonCodec emits a non-gRPC response content-type (`application/json`), breaking conforming gRPC clients

attila-lin:
## Summary
- `JsonCodec::CONTENT_TYPES` listed `application/json` before `application/grpc+json`, so generated handlers sent a non-gRPC content-type and grpc-go clients rejected the response.
- Reorder to match `JsonI64ToStringCodec` (same approach as #1191 by @kingcanfish).
- Also apply the rustfmt comment wraps that are currently failing `cargo fmt --check` on master, so CI can go green.

Closes #1190.
Supersedes #1191 with a minimal fix (content-type order only; no request-header echo).

## Test plan
- [ ] CI green on this PR
- [ ] Confirm `poem-grpc` tests pass


attila-lin:
Pushed a follow-up for the remaining `cargo fmt` hit in examples and a clippy `useless_borrows_in_formatting` on master that was failing `poem-openapi-derive`.

attila-lin:
Closing this in favor of keeping the original author-owned #1191. We'll ask for CI fixes there instead of landing a parallel maintainer PR.

